PR #87, opened 2024-11-08
Make useRateLimit and usePrometheus available from @graphql-hive/gateway

klippx:
### Description

This does not work

```ts
import { useRateLimit } from '@graphql-hive/gateway';
```

### Solution

Make `useRateLimit` exposed

### DIFF
--- a/.changeset/curly-socks-retire.md
+++ b/.changeset/curly-socks-retire.md
@@ -1,0 +1,5 @@
+---
+'@graphql-hive/gateway': patch
+---
+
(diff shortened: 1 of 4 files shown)


.changeset/curly-socks-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-hive/gateway': patch
```
ardatan, line 2:
```suggestion
'@graphql-hive/gateway': minor
```

PR #98, opened 2024-11-11
Bun support, E2E tests and Docker images

ardatan:
- [x] Avoid patching graphql-js for Bun

~~Blocked by https://github.com/ardatan/whatwg-node/pull/1814~~
- Compression plugin fails because of missing (De)CompressionStream impl in Bun, and this has been fixed in the mentioned PR.


enisdenjo (approved):

enisdenjo (changes requested):
Bun binary is not working when "graphql" is not available in the nearest node_modules.

enisdenjo:
There's also a bunch of warnings during `yarn install`.

enisdenjo (approved):

PR #104, opened 2024-11-12
Introduce Jest only for Leak tests

ardatan:


.github/workflows/ci.yml
```diff
@@ -17,7 +17,244 @@ jobs:
...
+        uses: ScribeMD/docker-cache@0.5.0
+        continue-on-error: true
+        with:
+          key: docker-images-${{ runner.os }}-${{ steps.hash-docker-images.outputs.result }}
+      - name: Bench
+        run: yarn e2e:bench
+
+  # --- START RELEASE ---
   snapshot:
+    needs: [e2e]
```
enisdenjo, line 257: I dont really like this. Means we have to wait for like ~10 mins to get a snapshot.

.github/workflows/ci.yml
```diff
@@ -29,6 +266,7 @@ jobs:
       githubToken: ${{ secrets.BOT_GITHUB_TOKEN }}
       npmToken: ${{ secrets.NPM_TOKEN }}
   stable:
+    needs: [bench]
```
enisdenjo, line 269: Why is this so random, somewhere you need bench somewhere e2es?

.github/workflows/ci.yml
ardatan, line 257:
```suggestion
```

ardatan: Removed

PR #105, opened 2024-11-12
Move Execution Runtime from Mesh repo

ardatan:


enisdenjo (commented):
Should we merge fusion-runtime and gateway-runtime?

ardatan:
> Should we merge fusion-runtime and gateway-runtime?

Not for now. Because gateway-runtime has HTTP stuff and fusion-runtime deals with the execution only.
We can think of refactoring later.

PR #177, opened 2024-11-24
Federated queries are not correctly resolved if one of subgraphs is unavailable

kroupacz:
Imagine a simple federated query:
```graphql
query {
  testNestedField {
    subgraph1 { # Everything inside this field is resolved by subgraph1
      testSuccessQuery {
        id
        email
        sub1
      }
    }
    subgraph2 { # Everything inside this field is resolved by subgraph2
      testSuccessQuery {
        id
        email
        sub2
      }
    }
  }
}
```

If subraph2 is unavailable, the query should be resolved as follows:
```ts
{
  testNestedField: {
    subgraph1: {
      testSuccessQuery: {
        id: 'user1',
        email: 'user1@example.com',
        sub1: true,
      },
    },
    subgraph2: {
      testSuccessQuery: null
    },
  },
}
```

In case of a subgrah2 outage, Apollo Gateway will resolve the above-mentioned query exactly as shown in the example. Unfortunately, GraphQL Yoga used as gateway does not work properly in this case and the result is also strangely affected by non/nullable **parent** fields even if the actual query **result has always nullable type**.

I didn't find the exact location where the problem occurs, but I would guess that it's somewhere in the `@graphql-tools/executor-http` package.

**Explanation of the tests which can be found in this PR.**
The first two test cases in `describe('EXPECTED RESPONSE - EVERYTHING OK')` when subgraph2 actually returns an error from its resolver return the expected response. These are just example tests of how it should work and result should be the same even if the subgrah2 is down.

The other 4 test cases in `describe('UNEXPECTED RESPONSE - FAILING TESTS')` when subgraph2 is actually unavailable always return a different and wrong response.

- test `subgraph2_Nullable.testErrorQuery - subgraph2 is unavailable`
  - unexpected response `data` and `errors` got completely lost somewhere along the way

- test `subgraph2_NonNullable.testErrorQuery - subgraph2 is unavailable`
  - unexpected response `data` and wrong error due to **NonNullable "parent"** field (GQL result type of `testErrorQuery` **is actually nullable**)

- test `subgraph2_Nullable.testErrorQuery only - subgraph2 is unavailable`
  - unexpected response `data`, but the error is **almost correct**. Only error `path` is incomplete and should be `['testNestedField', 'subgraph2_Nullable', 'testErrorQuery']`

- test `subgraph2_NonNullable.testErrorQuery only - subgraph2 is unavailable`
  - unexpected response `data`, but the error is **almost correct**. Only error `path` is incomplete and should be `['testNestedField', 'subgraph2_NonNullable', 'testErrorQuery']`

### DIFF
--- a/packages/federation/tests/unavailable-subgraph/yoga.gateway.test.ts
+++ b/packages/federation/tests/unavailable-subgraph/yoga.gateway.test.ts
@@ -1,0 +1,289 @@
+import {
+  FormattedExecutionResult,
+  GraphQLFormattedError,
+  versionInfo,
+} from 'graphql';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import { TestEnvironment } from './fixtures/TestEnvironment';
+
+const describeIf = (condition: boolean) =>
+  condition ? describe : describe.skip;
+
+describeIf(versionInfo.major >= 16)(
+  'Yoga gateway - subgraph unavailable',
+  () => {
+    let ctx: TestEnvironment;
+
+    beforeAll(async () => {
+      ctx = new TestEnvironment();
+      await ctx.start();
+    });
+
+    afterAll(async () => {
+      await ctx.stop();
+    });
+
+    const nullableQuery = /* GraphQL */ `
+      query {
+        testNestedField {
+          subgraph1_Nullable {
+            testSuccessQuery {
+              id
+              email
+              sub1
+            }
+          }
+          subgraph2_Nullable {
+            testErrorQuery {
+              id
+              email
+              sub2
+            }
+          }
+        }
+      }
+    `;
+
+    const expectedNullableData = {
+      testNestedField: {
+        subgraph1_Nullable: {
+          testSuccessQuery: {
+            id: 'user1',
+            email: 'user1@example.com',
+            sub1: true,
+          },
+        },
+        subgraph2_Nullable: {
+          testErrorQuery: null,
+        },
+      },
+    };
+
+    const nonNullableQuery = /* GraphQL */ `
+      query {
+        testNestedField {
+          subgraph1_NonNullable {
+            testSuccessQuery {
+              id
+              email
+              sub1
+            }
+          }
+          subgraph2_NonNullable {
+            testErrorQuery {
+              id
+              email
+              sub2
+            }
+          }
+        }
+      }
+    `;
+
+    const expectedNonNullableData = {
+      testNestedField: {
+        subgraph1_NonNullable: {
+          testSuccessQuery: {
+            id: 'user1',
+            email: 'user1@example.com',
+            sub1: true,
+          },
+        },
+        subgraph2_NonNullable: {
+          testErrorQuery: null,
+        },
+      },
+    };
+
+    const expectedUnavailableErrors: GraphQLFormattedError[] = [
+      {
+        message: expect.stringMatching(
+          /^connect ECONNREFUSED (127.0.0.1|::1):\d{4}$/,
+        ),
+        path: ['testNestedField'], // path should actually be ['testNestedField', 'subgraph2_Nullable | subgraph2_NonNullable', 'testErrorQuery'],
+      },
+    ];
+
+    describe('EXPECTED RESPONSE - EVERYTHING OK', () => {
+      it('subgraph2_Nullable.testErrorQuery - receive valid error if testErrorQuery throw an error', async () => {
+        const expectedErrors: GraphQLFormattedError[] = [
+          {
+            message: 'My original subgraph2 error!',
+            path: ['testNestedField', 'subgraph2_Nullable', 'testErrorQuery'],
+          },
+        ];
+
+        const response = await ctx.yoga.fetch('http://yoga/graphql', {
+          method: 'POST',
+          body: JSON.stringify({ query: nullableQuery }),
+          headers: { 'content-type': 'application/json' },
+        });
+
+        const result: FormattedExecutionResult = await response.json();
+
+        expect(response.status).toBe(200);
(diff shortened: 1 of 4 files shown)


ardatan:
Thanks for creating the PR, we'll look into it but it seems something is wrong with your setup.

kroupacz:
Hello @ardatan,
thanks for quick reply.
I pushed the fix for my `TestEnvironment.ts`. It should work now in your CI environment.
I would like to point out that my "failing tests" are written in such a way that they are not failing 😇, but are described in detail in comments where it should be clear what is wrong.

packages/federation/tests/unavailable-subgraph/yoga.gateway.test.ts
```diff
@@ -0,0 +1,289 @@
...
+        const response = await ctx.yoga.fetch('http://yoga/graphql', {
+          method: 'POST',
+          body: JSON.stringify({ query: nullableQuery }),
+          headers: { 'content-type': 'application/json' },
+        });
+
+        const result: FormattedExecutionResult = await response.json();
+
+        expect(response.status).toBe(200);
+        expect(result.data).toMatchObject(unexpectedData);
```
kroupacz, line 185: @ardatan This is interesting. If this test runs on Bun, it correctly receive `expectedNullableData` mentioned above. 🤔

kroupacz:
Hello @ardatan,
have you been able to investigate the issue yet?
Do you need any additional information which could help you solve the issue?
Thanks you in advance for your reply.
Sincerely,
Tomáš

ardatan:
After taking another look at the issue, I noticed few things;

If Subgraph 2 is down, subgraph2 field should be null with errors not `testSuccessQuery` one. 
Same for Subgraph 1 and subgraph1 field.

Because;
```json
{
  testNestedField: {
    subgraph1: {
      testSuccessQuery: {
        id: 'user1',
        email: 'user1@example.com',
        sub1: true,
      },
    },
    subgraph2: {
      testSuccessQuery: null
    },
  },
}
```

This means `subgraph2` field is available but `testSuccessQuery` field under of it is not.

kroupacz:
I understand your deduction, but I don't think it's correct in this case. If I run the same tests against a real Apollo Gateway, the result is correct in my opinion, or rather it behaves as the GraphQL specification says https://spec.graphql.org/draft/#sel-GAPHRPTPCAACEzBg6S

> If the field which experienced an error was declared as Non-Null, the null result will bubble up to the next nullable field

And the first non-nullable field is always `testSuccessQuery` or `testErrorQuery`.

However, if I run the tests against GraphQL Yoga (used as a gateway), it behaves differently and in my opinion incorrectly as you can see in the tests.

ardatan:
Let's think of an example outside Federation then;

We have a following schema;

```graphql
type Query {
  foo: Foo
}

type Foo {
  bar: Bar
}

type Bar {
  id: ID
}
```

and we have the following resolvers;

```ts
const resolvers = {
  Query: {
    foo: () => ({})
  },
  Foo: {
    bar: () => null,
  },
  Bar: {
    id: () => 'BAR_ID'
  }
}
```

What do you expect when you make the following request?

```graphql
{
  foo {
    bar {
      id
    }
  }
}
```

This one right? because `bar` is null, and there is no root value to put `id`.
```json
{
  "data": {
     "foo": {
        "bar": null
      }
  }
}
```

When we go back to our example, `subgraph2` is null and there is no object to add `testSuccessQuery`.

kroupacz:
Hello @dotansimha and @ardatan,
sorry for the late reply. @ardatan's answer makes sense. 👍 This is a problem for us because Apollo Gateway behaved differently in this, however we have already planned a modification to set each GraphQL schema field mapping the path to each subgraph as "nullable", which will solve our problem.

So I only have one problem left and that is with the test `subgraph2_Nullable.testErrorQuery - subgraph2 is unavailable` [here](https://github.com/graphql-hive/gateway/blob/eaeff3cd972b56496488b8f9bd3fea7fc7f4b9bb/packages/federation/tests/unavailable-subgraph/yoga.gateway.test.ts#L174).
If subgraph2 is unavailable I should get some ECONNREFUSED error in the response but now I only receive `subgraph2_Nullable: null` without any explanation.

dotansimha:
> @ardatan's answer makes sense. 👍 This is a problem for us because Apollo Gateway behaved differently in this, however we have already planned a modification to set each GraphQL schema field mapping the path to each subgraph as "nullable", which will solve our problem.

Yeah, I think we should not use Apollo Gateway as a reference here, as it's no longer maintained? I'm not sure if it was originated in a bug or intended behavior. 

> So I only have one problem left and that is with the test `subgraph2_Nullable.testErrorQuery - subgraph2 is unavailable` [here](https://github.com/graphql-hive/gateway/blob/eaeff3cd972b56496488b8f9bd3fea7fc7f4b9bb/packages/federation/tests/unavailable-subgraph/yoga.gateway.test.ts#L174). If subgraph2 is unavailable I should get some ECONNREFUSED error in the response but now I only receive `subgraph2_Nullable: null` without any explanation.

Yeah, @ardatan was able to reproduce this and we are testing a fix in https://github.com/graphql-hive/gateway/pull/929

kroupacz:
Hello @ardatan,
I just install `@graphql-tools/executor-http@1.3.2` where the fix https://github.com/graphql-hive/gateway/pull/929 should be included, but I believe it does not fix the above mentioned issue.
Please install the current yoga dependencies to this pull request and try run my test again. 🙏

kroupacz:
Hello @ardatan,
[here](https://github.com/graphql-hive/gateway/pull/950) you can find it.
If you have any questions, please don't hesitate to contact me.
Have a nice day.
Tomáš

PR #180, opened 2024-11-25
enhance: use the new explicit resource management in whatwg-node

ardatan:
Nothing changes on user end but it cleans up the gateway code by using the new feature from whatwg-node;
https://github.com/ardatan/whatwg-node/pull/1830


packages/executors/http/src/index.ts
```diff
@@ -91,96 +95,64 @@ export interface HTTPExecutorOptions {
    */
   print?: (doc: DocumentNode) => string;
   /**
-   * Enable [Explicit Resource Management](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management)
-   * @default false
+   * @deprecated Not used anymore
```
enisdenjo, line 98: Would be goot to explain shortly why.

packages/executors/http/tests/retry-timeout.test.ts
```diff
@@ -111,7 +111,7 @@ describe('Retry & Timeout', () => {
     expect(result).toMatchObject({
       errors: [
         {
-          message: expect.stringMatching(/timeout/),
+          message: expect.stringMatching(/out/),
```
enisdenjo, line 114: Just `out`? Kinda weird stringmatch. Is it because of case sensitivity or "time out" vs "timeout"?

enisdenjo (changes requested):
I'm loving it! Just two small remarks.

packages/executors/http/tests/retry-timeout.test.ts
ardatan: It is `timed out`  in Bun, and `timeout` in Node.

ardatan: Replaced it with better regexp

packages/executors/http/src/index.ts
ardatan: Done!

enisdenjo (approved):
Great stuff!

PR #204, opened 2024-11-27
opentelemetry => azure-monitor/application-insights

SvenHamers:
azure monitor / application insights support


ardatan:
That sounds interesting!
I made some changes in order to keep the plugin platform agnostic. So Azure exporter module will be loaded lazily. So non-Node enviroments won't break. Let me know if it sounds good!

PR #205, opened 2024-11-27
Handle provided fields and nested keys better

ardatan:
Refactor handling `@provides` directive


package.json
```diff
@@ -22,7 +22,9 @@
     "test": "vitest --project unit",
     "test:bun": "bun test --bail",
     "test:e2e": "vitest --project e2e",
-    "test:leaks": "cross-env \"LEAK_TEST=1\" jest --no-watchman --runInBand --detectOpenHandles --detectLeaks --logHeapUsage"
+    "test:leaks": "cross-env \"LEAK_TEST=1\" jest --no-watchman --runInBand --detectOpenHandles --detectLeaks --logHeapUsage",
+    "build:alpha": "yarn build && yarn pin-workspace-deps",
```
enisdenjo: Each package has a `prepack` script which will build only the package being released. So building all packages is completely unnecessary.

```suggestion
    "build:alpha": "yarn pin-workspace-deps",
```

package.json
```diff
@@ -22,7 +22,9 @@
...
+    "pin-workspace-deps": "node ./scripts/pin-workspace-deps.mjs"
```
enisdenjo: You should write the scripts in typescript, add them to the `include` in tsconfig and execute with tsx. This way everything is kept typesafe.

enisdenjo (changes requested):

.github/workflows/release.yml
```diff
@@ -6,7 +6,9 @@ on:
       - main
   pull_request:
 
-concurrency: ${{ github.workflow }}-${{ github.ref }}
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
enisdenjo, line 11: Not sure how safe it is to cancel a running release workflow. 🤔 

But I guess, even if the release did not complete altogether - the release CI that superseeds should.

ardatan: Reverted 👍

enisdenjo (approved):

PR #236, opened 2024-12-02
feat: AppDynamics integration plugin

### LINKED ISSUE #233: Create an AppDynamics plugin
# Context

AppDynamics is a Cisco technology for monitoring, trancing and error reporting.

It has automated Node instrumentation, and works with most used packages. In particular, it offers instrumentation of the `graphql` library. But since Mesh doesn't rely on it, it doesn't work automatically. We have to make a plugin to integrate with it.

# Solutions

I have identified 2 ways we could integrate with AppDynamics:
- use our OTEL plugin and configure AppDynamics to injest OTEL metrics
- implement a custom plugin using AppDynamics Node API.

## OTEL integration

This solution would have the huge advantage to require less work on our end, both in upfront development time and maintenance.

Based what is claimed in the documentation, it is supposed to "Just Work™️" after enabling the OTEL integration in the options. It is supposed to then correlate OTEL metrics and AppDynamics specific data and merge everything correctly.

The main drawback of this is that it requires to run an additional component. Users will have to setup an OTEL Collector, and configure it to export metrics to AppDynamics.

While the approach is a good idea and perhaps something we could explore afterward, I think the drawback of requiring a running OTEL Collector is too much to ask. Users going for this solution are probably doing so for it's ease of setup. And it can be difficult to meet this requirement in some env, like if the gateway is on the Edge.

## Custom plugin using Node agent API

The second solution is to create a plugin from scratch, relying on AppDynamics Node Agent API.

From my understanding, there is 2 main concepts that are important for us:
- Transactions: can be started manually (with an optional correlation information), represent an execution span.
	- It seems we can attach some data to it using "snapshots", not sure what it is yet.
	- Seems a bit tricky to use, will probably require a lot of testing. Namely, it seems you have to manually "re-join" a transaction each time you cross async boundaries.
- Exit Call: represents what they name "downstream" requests.
	- It can be whatever external services
	- It can be of any kind (HTTP, Database, ...)
	- That's how we will have to track our subgraph queries
	- Since we rely on `node:http` and `node:https` for our requests, it is supposed to be automatically handled
	- Otherwise, we will have to manually handle the creation of correlation information and propagate them through headers

The documentation is very sparse, so it will probably require a lot of trial and error to make it work.

They do not provide any testing tool, so it will probably be very hard to test this plugin.

They do not provide any types, we will have to write our own. Perhaps we could then contribute them back to the DefinitelyTyped repo.

# Propagation of transactions

In AppDynamics, correlation and propagation is supposed to be automatic. They instrument most used libraries to automatically add correlation info to outgoing calls, and detect them in ingoing calls (like HTTP requests).

We will probably have to manually check for those correlation information in a Yoga plugin so that transactions are correctly propagated in subgraphs if they are using Yoga.

Most of the work should be done automatically, it's just a matter of finding out how correlation data is transmitted (which HTTP header is used).

For other subgraph server library, they will have to either be automatically handled by AppDynamics (like graphql-http) or have a dedicated integration.

EmrysMyrddin:
fixes #233

### DIFF
--- a/packages/plugins/appdynamics/package.json
+++ b/packages/plugins/appdynamics/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@graphql-mesh/plugin-appdynamics",
-  "version": "1.3.11",
+  "version": "0.0.0",
   "type": "module",
   "repository": {
     "type": "git",
--- a/packages/plugins/appdynamics/package.json
+++ b/packages/plugins/appdynamics/package.json
@@ -14,7 +14,7 @@
   },
   "license": "MIT",
   "engines": {
-    "node": ">=18.0.0"
+    "node": ">=18.0.0 <23.0.0"
   },
   "main": "./dist/index.js",
   "exports": {
@@ -39,7 +39,6 @@
     "prepack": "yarn build"
   },
   "peerDependencies": {
-    "appdynamics": "^24.10.0",
     "graphql": "^15.9.0 || ^16.9.0"
   },
   "dependencies": {
@@ -54,7 +53,6 @@
     "tslib": "^2.4.0"
   },
   "devDependencies": {
-    "appdynamics": "^24.10.0",
     "graphql": "^16.9.0",
     "graphql-yoga": "^5.7.0",
     "pkgroll": "2.5.1"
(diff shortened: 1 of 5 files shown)


packages/plugins/appdynamics/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@graphql-mesh/plugin-appdynamics",
+  "version": "1.3.11",
```
ardatan, line 3:
```suggestion
  "version": "0.0.0",
```

ardatan:
Also it'd better to create a new changeset to trigger alpha releases

PR #313, opened 2024-12-12
feat: APQ between subgraph and gateway

### LINKED ISSUE #272: [Feature Request] APQ between GW and subgraphs
Follow APQ spec for the communication between subgraphs and the gateway;
https://www.apollographql.com/tutorials/caching-in-router/06-sending-apq-subgraphs

ardatan:
Closes #272 

Automatic Persisted Queries support for upstream requests

For HTTP Executor;
```ts
buildHTTPExecutor({
    // ...
    apq: true,
})
```

For Gateway Configuration;
```ts
    transportEntries: {
        '*': {
            options: {
                apq: true
            }
        }
    },
})
```

### DIFF
--- a/e2e/apq-subgraphs/gateway.config.ts
+++ b/e2e/apq-subgraphs/gateway.config.ts
@@ -1,0 +1,20 @@
+import { defineConfig } from '@graphql-hive/gateway';
+
+let fetchCnt = 0;
+export const gatewayConfig = defineConfig({
+  transportEntries: {
+    greetings: {
+      options: {
+        apq: true,
+      },
+    },
+  },
+  plugins: () => [
+    {
+      onFetch({ options }) {
+        fetchCnt++;
+        process.stdout.write(`fetch ${fetchCnt} ${options.body}\n`);
+      },
+    },
+  ],
+});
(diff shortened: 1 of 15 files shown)


e2e/apq-subgraphs/gateway.config.ts
```diff
@@ -0,0 +1,20 @@
+import { defineConfig } from '@graphql-hive/gateway';
+
+let fetchCnt = 0;
+export const gatewayConfig = defineConfig({
+  transportEntries: {
+    greetings: {
+      options: {
+        apq: true,
```
EmrysMyrddin, line 8: Transport entries is marked as an advanced feature today. Do we want to hide this behind this setting ?

Perhaps we can have a top level option instead ?

ardatan: They are not advanced. It is a way to configure the transport.

EmrysMyrddin: I don't think it's advanced, but that's what we are advertising in the documentation ^^'

And the way transport entries work is never actually explain. It's shortly explained in the Subscriptions page because it is needed to configure websocket/http callback subscription transports

ardatan: We should fix it then yes. I agree it is a bit low level but i think we should cover each feature individial rather than grouping them into one single transportEntries documentation. What do you think?

PR #322, opened 2024-12-12
feat: Retry and Timeout configuration for upstream requests

ardatan:
New Retry and Timeout plugins;

- Retry plugin: Retry a request if it fails

It respects the `Retry-After` HTTP header, [See more about this HTTP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After)

```ts
    upstreamRetry: {
        // The maximum number of retries to attempt.
        maxRetries: 3, // required
        // The maximum delay between retries in milliseconds.
        maxDelay: 1000, // default
        /**
         * A function that determines whether a response should be retried.
         * If the upstream returns `Retry-After` header, the response will be retried.
         */
        shouldRetry: response => response.status >= 500 || response.status === 429 // default
    }
    // or you can configure it by subgraph name
    upstreamRetry({ subgraphName }) {
        if (subgraphName === 'my-rate-limited-subgraph') {
            return {
                maxRetries: 3,
            }
        }
    }
})
```

- Timeout plugin: Timeout a request if it takes too long

```ts
    // The maximum time in milliseconds to wait for a response from the upstream.
    upstreamTimeout: 1000, // required
    // or you can configure it by subgraph name
    upstreamTimeout({ subgraphName }) {
        if (subgraphName === 'my-slow-subgraph') {
            return 1000;
        }
    }
})
```


packages/runtime/src/plugins/useUpstreamRetry.ts
```diff
@@ -0,0 +1,127 @@
...
+   * @default 1000
+   */
+  retryDelay?: number;
+  /**
+   * A function that determines whether a response should be retried.
+   * If the upstream returns `Retry-After` header, the response will be retried.
+   *
+   * @default (response) => response.status >= 500 || response.status === 429
+   */
+  shouldRetry?: (response: Response) => boolean;
```
EmrysMyrddin, line 23: We should probably have a `payload: { response: Response }` here, so that we can easily add other metadata, such as the execution context, the subgraph name, etc...

packages/runtime/src/plugins/useUpstreamRetry.ts
```diff
@@ -0,0 +1,127 @@
...
+                          const dateTime = new Date(retryAfterHeader).getTime();
+                          if (!isNaN(dateTime)) {
+                            retryAfterSeconds = dateTime - requestTime;
+                          }
+                        }
+                      }
+                      let currentRetryDelay: number | undefined;
+                      if (retryAfterSeconds) {
+                        currentRetryDelay = retryAfterSeconds * 1000;
+                      } else if (shouldRetry(response)) {
```
EmrysMyrddin, line 85: You always retry if the retry header is present ? Without checking for `shouldRetry` ?

packages/runtime/src/plugins/useUpstreamRetry.ts
```diff
@@ -0,0 +1,127 @@
...
+                  );
+                } catch (e) {
+                  if (retries < 0) {
+                    throw e;
+                  }
+                  return retry();
+                }
+              }
+              return retry();
+            },
```
EmrysMyrddin, line 115: Why is this duplicated ? You can't just let the error buble up to the catch ?

ardatan: `mapMaybePromise` doesn't handle sync errors

packages/runtime/src/plugins/useUpstreamRetry.ts
```diff
@@ -0,0 +1,127 @@
...
+  options: RequestInit;
+  context: GatewayContext;
+  subgraphName?: string;
+}
+
+export type UpstreamRetryPluginOptions =
+  | UpstreamRetryOptions
+  | ((payload: UpstreamRetryPayload) => UpstreamRetryOptions | undefined);
+
+export function useUpstreamRetry<TContext extends Record<string, any>>(
```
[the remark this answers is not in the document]
ardatan: Reimplemented in agnostic way!

packages/runtime/src/plugins/useUpstreamRetry.ts
ardatan: You're right! Fixed.

.changeset/witty-candles-whisper.md
```diff
@@ -0,0 +1,49 @@
...
+    upstreamRetry: {
+        // The maximum number of retries to attempt.
+        maxRetries: 3, // required
+        // The maximum delay between retries in milliseconds.
+        maxDelay: 1000, // default
+        /**
+         * A function that determines whether a response should be retried.
+         * If the upstream returns `Retry-After` header, the response will be retried.
+         */
+        shouldRetry: response => response.status >= 500 || response.status === 429 // default
```
EmrysMyrddin, line 23:
```suggestion
        shouldRetry: ({ response }) => response.status >= 500 || response.status === 429 // default
```

packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -1002,6 +1004,14 @@ export function createGatewayRuntime<
     extraPlugins.push(usePropagateHeaders(config.propagateHeaders));
   }
 
+  if (config.upstreamRetry) {
+    extraPlugins.push(useUpstreamRetry(config.upstreamRetry));
+  }
+
+  if (config.upstreamTimeout) {
+    extraPlugins.push(useUpstreamTimeout(config.upstreamTimeout));
+  }
```
EmrysMyrddin, line 1013: We should probably invert the order of plugins.

Currently, the timeout plugin comes after the retry plugin. It means the executor with a timeout will wrap the executor with the retry.

We probably want the other way around, the retry executor should wrap the timeout executor. This way, a timed out request will be retryable.

```suggestion
  if (config.upstreamTimeout) {
    extraPlugins.push(useUpstreamTimeout(config.upstreamTimeout));
  }
  
  if (config.upstreamRetry) {
    extraPlugins.push(useUpstreamRetry(config.upstreamRetry));
  }
```

packages/runtime/src/plugins/useUpstreamRetry.ts
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * The maximum number of retries to attempt.
+   */
+  maxRetries: number;
+  /**
+   * The delay between retries in milliseconds.
+   * If the upstream returns `Retry-After` header, the delay will be the value of the header.
+   * @default 1000
+   */
+  retryDelay?: number;
```
EmrysMyrddin, line 23: I didn't thought about it, but when it comes to retry policies, a backoff is often something important.

Like a simple exponential backoff, to avoid spamming the upstream if it is already having a hard time responding to requests :-P

Perhaps the retryDelay can be a function ? This way the user can define it's own backoff strategy.
Our default implementation could be to search for `Retry-After` header, and fallback to exponential backoff if not present ?

ardatan: That's what it does actually. It first look for the header if not present, it uses `retryDelay`

.changeset/witty-candles-whisper.md
```diff
@@ -0,0 +1,49 @@
...
+    upstreamRetry: {
+        // The maximum number of retries to attempt.
+        maxRetries: 3, // required
+        // The maximum delay between retries in milliseconds.
+        maxDelay: 1000, // default
+        /**
+         * A function that determines whether a response should be retried.
+         * If the upstream returns `Retry-After` header, the response will be retried.
+         */
+        shouldRetry: ({ response }) => response.status >= 500 || response.status === 429 // default
```
EmrysMyrddin, line 23: The response can be null, and it's not actually the real default implementation

```suggestion
        shouldRetry: ({ response }) => response?.status >= 500 || response?.status === 429
```

packages/runtime/src/plugins/useUpstreamRetry.ts
EmrysMyrddin: Yes, but I mean having a `retryDelay` that is not a fixed value. `retryDelay` should increase after each try. Something like `retryDelay *= 2`

.changeset/witty-candles-whisper.md
```diff
@@ -0,0 +1,49 @@
...
+
+It respects the `Retry-After` HTTP header, [See more about this HTTP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After)
+
+```ts
+export const gatewayConfig = defineConfig({
+    upstreamRetry: {
+        // The maximum number of retries to attempt.
+        maxRetries: 3, // required
+        // The maximum delay between retries in milliseconds.
+        maxDelay: 1000, // default
```
EmrysMyrddin, line 18:
```suggestion
        retryDelay: 1000, // default
```

packages/runtime/src/plugins/useUpstreamRetry.ts
```diff
@@ -0,0 +1,174 @@
...
+                        executionResult,
+                        response,
+                      })
+                    ) {
+                      currentRetryDelay ||= retryDelay;
+                      return new Promise((resolve) => {
+                        const timeout = setTimeout(() => {
+                          timeouts.delete(timeout);
+                          resolve(retry());
+                        }, retryDelay);
```
EmrysMyrddin, line 134: The `retryDelay` is used instead of the computed `currentRetryDelay` variable.

```suggestion
                        }, currentRetryDelay);
```

packages/runtime/src/plugins/useUpstreamRetry.ts
ardatan: Anyway what do you think now?

packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -93,6 +93,8 @@ import { usePropagateHeaders } from './plugins/usePropagateHeaders';
 import { useRequestId } from './plugins/useRequestId';
 import { useSubgraphExecuteDebug } from './plugins/useSubgraphExecuteDebug';
 import { useUpstreamCancel } from './plugins/useUpstreamCancel';
+import { useUpstreamRetry } from './plugins/useUpstreamRetry';
+import { useUpstreamTimeout } from './plugins/useUpstreamTimeout';
```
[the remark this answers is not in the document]
ardatan: Done!

packages/runtime/src/plugins/useUpstreamTimeout.ts
```diff
@@ -0,0 +1,55 @@
...
+          options.signal.addSignals([timeoutSignal]);
+        } else {
+          const signals = [timeoutSignal];
+          if (options.signal) {
+            signals.push(options.signal);
+          }
+          options.signal = abortSignalAny(signals);
+        }
+      }
+    },
```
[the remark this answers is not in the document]
ardatan: Done!

packages/runtime/src/plugins/useUpstreamRetry.ts
ardatan: Done!

packages/runtime/src/plugins/useUpstreamRetry.ts
```diff
@@ -87,11 +87,13 @@ export function useUpstreamRetry<TContext extends Record<string, any>>(
         if (maxRetries > 0) {
           setExecutor(function (executionRequest: ExecutionRequest) {
             let retries = maxRetries + 1;
+            let currentMinRetryDelay = minRetryDelay;
             let executionResult: MaybeAsyncIterable<ExecutionResult>;
             function retry(): MaybePromise<
               MaybeAsyncIterable<ExecutionResult>
             > {
               retries--;
+              currentMinRetryDelay += minRetryDelay;
```
EmrysMyrddin: This gives us linear backoff. The standard is more leaning towards exponential backoff, with a rate of something between 1 and 2.
```suggestion
              currentMinRetryDelay = currentMinRetryDelay * 2;
```

.changeset/witty-candles-whisper.md
```diff
@@ -0,0 +1,49 @@
...
+         */
+        shouldRetry: ({ response }) => response?.status >= 500 || response?.status === 429
+    }
+    // or you can configure it by subgraph name
+    upstreamRetry({ subgraphName }) {
+        if (subgraphName === 'my-rate-limited-subgraph') {
+            return {
+                maxRetries: 3,
+            }
+        }
```
EmrysMyrddin, line 31: Perhaps add a default number of retries ? Otherwise it means only the rate limited subgraph is the only one that get retries :-P 
```suggestion
        }
        return { maxRetries: 10 }
```

EmrysMyrddin (approved):

PR #323, opened 2024-12-12
feat: pass extensions to delegated request

lchudinov:
# Pass extensions to delegegated request

### DIFF
--- a/packages/delegate/src/createRequest.ts
+++ b/packages/delegate/src/createRequest.ts
@@ -175,6 +175,7 @@ export function createRequest({
     context,
     info,
     operationType: targetOperation,
+    extensions: context.params?.extensions,
   };
 }
 


ardatan:
Could you tell more about this PR?
Because this might cause a leak of extensions to the upstream requests? I think it should be opt-in.

lchudinov:
Hello @ardatan,

When I send a GraphQL request with extensions to the stitched GraphQL schema, for example:  
```json
{
  "query": "query MyQuery { myQuery }",
  "operationName": "MyQuery",
  "extensions": { "debug": true }
}
```
the `extensions` is missing when the delegated request reaches the target schema.

lchudinov:
I figured out that I can forward the extensions if I create a custom executor.

PR #328, opened 2024-12-13
[WIP] Refactor open telemetry plugin

darren-west:
### Summary
This is a refactor to make the opentelemetry plugin in the hive gateway conform closer to the opentelemetry api spec specifically around how context is propagated within the node sdk.

The goal is to allow the gateway to take advantage of the instrumentations available in the otel libraries, for example http, dns, grpc etc. The big benefit of this is it allows custom plugins to benefit from the instrumentations which is a requirement that we have, where we create a grpc request within the context phase of GraphQL. 

I put some effort into removing potentially sensitive data from traces, for example error messages. Im not sure on the best approach perhaps an option could be passed in to support this if its a requirement for others. 

In otel for node it uses async local storage to wrap functions to propagate the context, this PR introduces the wrapping of the different phases of GraphQL so that any span has the correct trace parent. This can be seen when using http instrumentation, because the execute phase is wrapped using api.with the subsequent fetch will have the correct trace parent.    

### Questions to answer.
- [ ] The open telemetry instrumentations predominantly work by patching, which means that the bootstrapping code needs to run before packages are imported. An example is the patching of gRPC or fetch. This makes it not fit with running this logic in a plugins init phase as its to late. 
- [ ] How do we configure the bootstrapping of otel within the hive-gateways configuration.
- [ ] Testing this as a gateway, whats the best approach for instance how can I test the onSubgraphExecute function?

TODO:
- [ ] Refactor and add tests, reintroducing existing features.
- [ ] Figure out where best to run the bootstrapping of the otel library.


packages/plugins/opentelemetry/src/setup.ts
```diff
@@ -0,0 +1,67 @@
+// This must be run first. Node uses patching on implementations to inject telemetry so must run before the packages
```
darren-west, line 1: This would not live inside the plugin as it is server setup for otel. 
The way otel is setup is by patching of libraries such as fetch, for this reason it needs to run outside of the plugin and be one imported before most packages.

packages/plugins/opentelemetry/src/setup.ts
```diff
@@ -0,0 +1,67 @@
...
+    ATTR_SERVICE_NAME,
+    ATTR_TELEMETRY_SDK_LANGUAGE, ATTR_TELEMETRY_SDK_VERSION
+} from "@opentelemetry/semantic-conventions";
+import {OTLPTraceExporter} from "@opentelemetry/exporter-trace-otlp-grpc";
+
+const contextManager = new AsyncLocalStorageContextManager().enable();
+api.context.setGlobalContextManager(contextManager);
+
+const exporter = new OTLPTraceExporter({
+    url: "todo:configuration",
```
darren-west, line 25: We would need to drive this from configuration. I have done this within our gateway implementation but that has a different approach to the hive-gateway to config so Im not sure of the best approach.

packages/plugins/opentelemetry/src/utils.ts
```diff
@@ -0,0 +1,45 @@
+import {ASTNode, DocumentNode, Kind, visit} from "graphql/index";
```
darren-west, line 1: todo: test this

packages/plugins/opentelemetry/tests/useOpenTelemetry.spec.ts
```diff
@@ -1,131 +1,124 @@
...
+                counter(count: Int!): Int!
             }
-          `,
-          resolvers: {
+        `,
+        resolvers: {
             Query: {
-              hello: () => 'World',
+                ping: () => {
+                    expect(api.context.active()).not.toEqual(api.ROOT_CONTEXT); // proves that the context is propagated
```
darren-west, line 37: Might want to figure out a better way to test this.

dotansimha:
Hi @darren-west , we've just merged https://github.com/graphql-hive/gateway/pull/532 , @EmrysMyrddin did amazing work on refactoring the internals of the plugin system and gw core, so now `instrumentation` workflows are easier to trace (https://github.com/graphql-hive/gateway/pull/706).

The new OTEL plugin relies on the core instrumentation, so OTEL spans are now fully correlated with the gw runtime - also includes fixes for how we handle terminations, retries, batched requests, request otel context and many other areas where the previous implementation failed to deliver. 

@EmrysMyrddin took a lot of inspiration for this PR, thanks @darren-west ! 

Upcoming PRs are going to address most of the open issues related to OTEL. We are also experimenting with new v2 of otel-js sdk (https://github.com/graphql-hive/gateway/pull/875). 

I'm going to close this one, @darren-west please let me know if we missed anything, we can create follow-up issues and plan this in our next iteration on the OTEL plugin 🙏

Issue #310, opened 2024-12-10
Self-hosted usage reporting not working


jlambynz:
When running the gateway using `supergraph`, adding configuration to support usage reporting to a self hosted instance, Hive is still trying to use the cloud endpoint.

Given a config something like:
```
export const gatewayConfig = defineConfig({
  supergraph: {
    type: "hive",
    endpoint: "https://self-hosted.com/artifacts/..."
    key: 123
  },
  reporting: {
    type: "hive",
    token: "123",
    selfHosting: {
      applicationUrl: "https://self-hosted.com",
      graphqlEndpoint: "https://self-hosted.com/graphql",
      usageEndpoint: "https://self-hosted.com/usage",
    },
  },
```

The gateway doesn't use the `reporting.selfHosting` options and defaults to targeting `https://app.graphql-hive.com/graphql`:
![image](https://github.com/user-attachments/assets/e19f501c-d8ba-48d3-8b15-b85f0b84e0fe)

From what I can tell, `reporting` configuration is never passed and instead a basic reporting object is constructed only consisting of `type` and `token`: https://github.com/graphql-hive/gateway/blob/main/packages/gateway/src/commands/supergraph.ts#L157-L164

I notice in the [documentation ](https://the-guild.dev/graphql/hive/docs/self-hosting/get-started) it mentions not all functionality is available in the self-hosted version, including `usage-estimation`, but it's not clear to me if that includes usage reporting in general. Should this configuration work for the self hosting?

jlambynz:
@ardatan Not immediately confident I can easily create a reproduction but I'll review what you put together and see if that gleans any more information

In the interim, just to clarify as it seems I may have caused confusion, I'm not having any issues passing the token, it's about specifying a usage endpoint that isn't the default `https://app.graphql-hive.com/graphql`. In the config I specify `selfHosting.usageEndpoint` to point to our hosted instance, but the Hive CLI continues to try send usage to `https://app.graphql-hive.com/graphql`.

The link to the code I was referencing was me wondering if the `selfHosting` section of the config should be provided there? I did a test run locally where I changed that part of the code to

```
if (hiveRegistryToken) {
        ctx.log.info(`Configuring Hive registry reporting`);
        registryConfig = {
          reporting: {
            type: 'hive',
            token: hiveRegistryToken,
            ...loadedConfig.reporting // <----- added this line
          },
        };
      }
```

With this change, the Hive CLI started sending usage reports to our hosted instance

[...]

PR #342, opened 2024-12-16
Respect CLI arguments and configuration when both provided

ardatan:
Closes https://github.com/graphql-hive/gateway/issues/310

User can provide the token in the CLI arguments, and have some registry configuration in \`gateway.config\`

Since `token` can be provided in the CLI arguments, it should not be required in the configuration.

## [Extra but needed] : Fixes in Test suite

Besides the changes in the code, I also saw that \`bin\` serve runner was not there and the tests were never running with the binary.
So I fixed it by adding `bin` serve runner implementation in the test suite.
I also found that `process.env` is not passed to the Docker runners as we do for other runners.

But I couldn't figure out why `args` not passed to the Docker runners, I need @enisdenjo 's help there because we test `args` here for Docker as in the original issue.

### DIFF
--- a/packages/gateway/src/commands/supergraph.ts
+++ b/packages/gateway/src/commands/supergraph.ts
@@ -249,7 +249,9 @@ export const addCommand: AddCommand = (ctx, cli) =>
         process.exit(1);
       }
       return runSupergraph(ctx, config);
-    });
+    })
+    .allowUnknownOption(process.env.NODE_ENV === 'test')
+    .allowExcessArguments(process.env.NODE_ENV === 'test');
 
 export type SupergraphConfig = GatewayConfigSupergraph & GatewayCLIConfig;
 
--- a/packages/gateway/src/commands/supergraph.ts
+++ b/packages/gateway/src/commands/supergraph.ts
@@ -158,6 +158,7 @@ export const addCommand: AddCommand = (ctx, cli) =>
         ctx.log.info(`Configuring Hive registry reporting`);
         registryConfig = {
           reporting: {
+            ...loadedConfig.reporting,
             type: 'hive',
             token: hiveRegistryToken,
           },
--- a/packages/gateway/src/config.ts
+++ b/packages/gateway/src/config.ts
@@ -53,6 +53,7 @@ export async function loadConfig<
       if (exists) {
         !opts.quiet &&
           opts.log.info(`Found default config file ${absoluteConfigPath}`);
+        const importUrl = pathToFileURL(absoluteConfigPath).toString();
         const module = await import(importUrl);
         importedConfig = Object(module).gatewayConfig || null;
         if (!importedConfig) {
--- a/packages/gateway/src/config.ts
+++ b/packages/gateway/src/config.ts
@@ -52,6 +52,7 @@ export async function loadConfig<
         .catch(() => false);
       if (exists) {
         !opts.quiet &&
+          opts.log.info(`Found default config file ${absoluteConfigPath}`);
         const module = await import(importUrl);
         importedConfig = Object(module).gatewayConfig || null;
         if (!importedConfig) {
--- a/packages/gateway/src/config.ts
+++ b/packages/gateway/src/config.ts
@@ -52,10 +52,6 @@ export async function loadConfig<
         .catch(() => false);
       if (exists) {
         !opts.quiet &&
-          opts.log.info(`Found default config file ${absoluteConfigPath}`);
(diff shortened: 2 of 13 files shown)


ardatan:
CLI Arguments reference -> https://the-guild.dev/graphql/hive/docs/api-reference/gateway-cli

We also have CLI configuration reference but it doesn't even mention about reporting;
https://the-guild.dev/graphql/hive/docs/api-reference/gateway-config

But instead we have this page for the configuration API for reporting
https://the-guild.dev/graphql/hive/docs/gateway/usage-reporting

And it is not clear how to pass other options like `selfHosting` :/

packages/runtime/src/types.ts
```diff
@@ -185,8 +185,6 @@ export interface GatewayHiveCDNOptions {
 export interface GatewayHiveReportingOptions
   extends Omit<YamlConfig.HivePlugin, 'experimental__persistedDocuments'> {
   type: 'hive';
-  /** GraphQL Hive registry access token. */
-  token: string;
```
enisdenjo, line 189: There is no CLI or CLI arguments in the runtime. The `token` is still required in the runtime.

packages/gateway/src/config.ts
```diff
@@ -54,6 +54,8 @@ export async function loadConfig<
         !opts.quiet &&
           opts.log.info(`Found default config file ${absoluteConfigPath}`);
         const importUrl = pathToFileURL(absoluteConfigPath).toString();
+        !opts.quiet &&
+          opts.log.info(`Loading config file at path ${importUrl}`);
```
enisdenjo, line 58: There's already a similar enough log above, why have two?
```suggestion
```

enisdenjo (changes requested):

packages/gateway/src/config.ts
ardatan: I wanted to see the import path which is created by `pathToFileURL` but I can remove it.

packages/runtime/src/types.ts
ardatan: Then we need to make it optional for CLI only?

Urigo:
> CLI Arguments reference -> [the-guild.dev/graphql/hive/docs/api-reference/gateway-cli](https://the-guild.dev/graphql/hive/docs/api-reference/gateway-cli)
> 
> We also have CLI configuration reference but it doesn't even mention about reporting; [the-guild.dev/graphql/hive/docs/api-reference/gateway-config](https://the-guild.dev/graphql/hive/docs/api-reference/gateway-config)
> 
> But instead we have this page for the configuration API for reporting [the-guild.dev/graphql/hive/docs/gateway/usage-reporting](https://the-guild.dev/graphql/hive/docs/gateway/usage-reporting)
> 
> And it is not clear how to pass other options like `selfHosting` :/

Thanks!
Let's improve the docs around these as part of this PR (I understand its on a different repo...)

packages/gateway/src/commands/supergraph.ts
```diff
@@ -249,7 +250,9 @@ export const addCommand: AddCommand = (ctx, cli) =>
         process.exit(1);
       }
       return runSupergraph(ctx, config);
-    });
+    })
+    .allowUnknownOption(process.env.NODE_ENV === 'test')
```
dotansimha, line 254: Can you please elaborate on this change?

dotansimha (approved):
LGTM. Let's wait for @enisdenjo confirmation. 
@ardatan , do we need docs updates for this change? I

packages/gateway/src/commands/supergraph.ts
ardatan: When running tests, `NODE_ENV` passed as `test`, and this allows us to accept extra CLI arguments.
We need this but why?
Because during tests random ports are used and the CLI needs to know which ports are used for which services.
This was missing for GW CLI but only available for Mesh Compose. So in order to have a test self hosting service, we need to pass our test environment's port to the GW CLI.

ardatan:
Actually this is a PR for fixing a bug more than changing it. But I added docs for Hive Client reference here to make it clear for users need to use more features from Hive Console.

This is the PR improving docs for that purpose.
@dotansimha 
https://github.com/graphql-hive/console/pull/6145

packages/runtime/src/types.ts
ardatan: How about now? @enisdenjo

enisdenjo (approved):

PR #351, opened 2024-12-18
`@cacheControl` support tests, fixes...

ardatan:
Today if you `@cacheControl` directive in your subgraphs with Apollo Server, it sends cache control headers. And this is supported by http cache plugin. You don't need `@composeDirective` etc.

If you use those directives without ApolloServer, you can give the control to the gateway using Response Caching plugin but in order to have those directives in your gateway, you have to add the directive using `@composeDirective` so the supergraph can have it and response caching plugin can pick it up and use it.

This PR fixes the following issue when the directive definitions got lost while handling the supergraph using `@graphql-tools/federation`.
It also adds tests for both the cases mentioned above.

## ~Needs to be documented on the website!!!~(Done!)
See the conversation below with @kamilkisiela to make sure to get what is documented!

After merging this PR, we need to update the docs with the two cases. And we need to mention http caching plugin can be used with the subgraph services returning cache control headers on its own and no `@composeDirective`.
If the subgraph is expected to control gateway's caching, response caching should be added and `@composeDirective` should be added for `@cacheControl`;

```ts
extend schema
  @link(
    url: "https://specs.apollo.dev/federation/v2.1"
    import: ["@composeDirective"]
  )
  @link(url: "https://the-guild.dev/mesh/v1.0", import: ["@cacheControl"])
  @composeDirective(name: "@cacheControl")
  ```

### DIFF
--- a/.changeset/tricky-paws-walk.md
+++ b/.changeset/tricky-paws-walk.md
@@ -1,0 +1,5 @@
+---
+'@graphql-tools/federation': patch
+---
+
+Keep the custom directives(using @composeDirective) from the supergraph, in the unified schema served by the gateway should keep it.
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -6,10 +6,7 @@ import { defineConfig } from 'vitest/config';
 // packages as per the Node resolution spec.
 //
 // Vite will process inlined modules.
-const inline = [
-  /@graphql-mesh\/.*/,
-  /@omnigraph\/.*/,
-];
+const inline = [/@graphql-mesh\/.*/, /@omnigraph\/.*/];
 
 export default defineConfig({
   plugins: [tsconfigPaths()],
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -7,11 +7,7 @@ import { defineConfig } from 'vitest/config';
 //
 // Vite will process inlined modules.
 const inline = [
-  /@graphql-mesh\/utils/,
-  /@graphql-mesh\/runtime/,
-  /@graphql-mesh\/fusion-composition/,
-  /@graphql-mesh\/plugin-hive/,
-  /@graphql-mesh\/transport-rest/,
+  /@graphql-mesh\/.*/,
   /@omnigraph\/.*/,
 ];
 
--- a/examples/apq-subgraphs/package-lock.json
+++ b/examples/apq-subgraphs/package-lock.json
@@ -7,7 +7,7 @@
       "name": "@example/apq-subgraphs",
       "dependencies": {
         "@apollo/server": "^4.11.2",
-        "@graphql-hive/gateway": "^1.7.6",
+        "@graphql-hive/gateway": "^1.7.7",
         "@graphql-mesh/compose-cli": "^1.2.13",
         "graphql": "^16.9.0",
         "tslib": "^2.8.1"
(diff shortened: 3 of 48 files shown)

### CODE REVIEW

**Comment by kamilkisiela** [comment]
When somebody runs `hive schema:publish` and points to a schema with `@cacheControl` directive then it won't be passed to supergraph SDL. Do I understand correctly that `mesh-compose` is needed to print the SDL of the subgraph?

**Comment by ardatan** [comment]
> When somebody runs hive schema:publish and points to a schema with @cacheControl directive then it won't be passed to supergraph SDL.

You know `@cacheControl` is not part of federation spec. So `@composeDirective` should be used together with the directive definition like this;
```graphql
extend schema
  @link(
    url: "https://specs.apollo.dev/federation/v2.1"
    import: ["@composeDirective"]
  )
  @link(url: "https://the-guild.dev/mesh/v1.0", import: ["@cacheControl"])
  @composeDirective(name: "@cacheControl")
```

If they use mesh-compose, it respects `@cacheControl` and automatically generates `@composeDirective`.
But if they don't use it, they have to add these `composeDirective` definition for `cacheControl` on their own.

kamilkisiela:
Yeah, that's what I asked about. We need to make sure it's well documented

ardatan:
> Yeah, that's what I asked about. We need to make sure it's well documented

Yes 💯

vitest.config.ts
```diff
@@ -6,14 +6,7 @@ import { defineConfig } from 'vitest/config';
...
 // Vite will process inlined modules.
-const inline = [
-  /@graphql-mesh\/utils/,
-  /@graphql-mesh\/runtime/,
-  /@graphql-mesh\/fusion-composition/,
-  /@graphql-mesh\/plugin-hive/,
-  /@graphql-mesh\/transport-rest/,
-  /@omnigraph\/.*/,
-];
+const inline = [/@graphql-mesh\/.*/, /@omnigraph\/.*/];
```
ardatan, line 9: This issue again but this change fixes it for good. But maybe it is wrong @enisdenjo Wdyt?

Urigo:
@ardatan can you add documentation of this as part of this PR?

Urigo:
also, how does Apollo's gateways handle this?

ardatan:
> also, how does Apollo's gateways handle this?
If I don't miss anything ( @kamilkisiela can correct me if I'm wrong ) , it respects cache-control headers as our http cache plugin does since `@cacheControl` is not available in the supergraph without `@composeDirective`;
This is my reference; https://www.apollographql.com/docs/apollo-server/performance/caching

Urigo:
@ardatan I would like to review the docs and this together, as it will help me understand everything before I review

about Apollo, can you create an example and test how it actually behaves?
maybe a side by side example - one with Apollo Router and another with Hive Gateway?
Would also help me understand and might also help with docs

kamilkisiela:
Apollo Router caches only the entity calls (to my knowledge) and it respects `cache-control`.
It supports global ttl (per subgraph) - as it obviously don't know the entity types it exposes.
It does not cache entire response emitted to the client.

.changeset/tricky-paws-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-tools/federation': patch
+---
+
+Keep the custom directives(using @composeDirective) from the supergraph, in the unified schema served by the gateway should keep it.
```
dotansimha, line 5: Do we need to link here to some issue?

ardatan: It is a feedback on Slack.

ardatan:
@Urigo There is no special configuraiton needed for Apollo Gateway/Router in this case but there are two options in Hive Gateway as documented here;
https://github.com/graphql-hive/console/pull/6161

Urigo:
Waiting for improvements on the docs before approving

Urigo (approved):

PR #357, opened 2024-12-19
fix: setting the default polling interval correctly

ardatan:
Fix the bug on setting the default polling interval to 10 seconds
So by default, the gateway will poll the schema every 10 seconds, and update the schema if it has changed.

This PR also contains improvements on logging about polling

### DIFF
--- a/packages/runtime/src/createGatewayRuntime.ts
+++ b/packages/runtime/src/createGatewayRuntime.ts
@@ -662,7 +662,12 @@ export function createGatewayRuntime<
 
       if (!isDynamicUnifiedGraphSchema(config.supergraph)) {
         // no polling for static schemas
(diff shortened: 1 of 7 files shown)


packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -1087,10 +1092,6 @@ function isDynamicUnifiedGraphSchema(
     return false;
   }
   if (typeof schema === 'string') {
-    if (isValidPath(schema) && !isUrl(String(schema))) {
```
ardatan, line 1090: Even supergraphs on the file system should be polled

dotansimha (approved):

ardatan:
@dotansimha Documentation update

PR #379, opened 2024-12-25
Convert E2Es to examples

enisdenjo:
- Transforms an E2E test to a runnable example
  - Never worry about writing examples, write E2Es and let the `@internal/examples` converter do the rest
- The converter makes sure examples work by running them and checking whether gateway is healthy
- Runs in CI
  - Examples are always be up-to-date with the latest versions
  - Examples are continuously tested through E2Es and by the converter itself which makes sure the gateway is healthy after running
  - As we cut new versions, examples will update too
- Plug-n-Play for [Codesandbox Devboxes](https://codesandbox.io/docs/learn/devboxes/overview)
  - Installs with `npm i`
  - Intentionally outside workspace (each example will have its lockfile)
  - Defines [setup and tasks](https://codesandbox.io/docs/learn/devboxes/task) for the devbox and tests them
  - When viewing an example on github.com, just append "box" to "github" (githubbox.com) and you'll get the example running in a CodeSandbox Devbox
    - Run `extra-field` example from this branch with [github**box**.com/graphql-hive/gateway/tree/e2e2example/examples/extra-fields](https://githubbox.com/graphql-hive/gateway/tree/e2e2example/examples/extra-fields)

### TODO

- [x] Warn that examples are auto-generated and not to be modified directly
- [x] Container Sandboxes are deprecated, use [Codesandbox Devboxes](https://codesandbox.io/docs/learn/devboxes/overview) instead
- [x] Support `createExampleSetup`
- [x] Convert rest of relevant E2Es
- [ ] Run containers used for E2Es (do Codesandbox Deboxes even support it?)


dotansimha (approved):

.github/workflows/examples.yml
```diff
@@ -0,0 +1,95 @@
...
+        id: commit
+        run: echo "message=$(git show -s --format=%s)" >> "$GITHUB_OUTPUT"
+
+  convert:
+    needs: [head-commit]
+    if: "${{needs.head-commit.outputs.message != 'docs(examples): converted from e2es'}}"
+    strategy:
+      matrix:
+        e2e:
+          - extra-fields
```
ardatan, line 32: This will create workflows for each example which might cause a bigger queue. But instead maybe we can run this once and configure converted e2es in a file, no?

Urigo:
very cool

PR #381, opened 2024-12-26
Fix HMAC plugin and single place to serialize upstream execution request in different components

ardatan:
Like HMAC Upstream Signature plugin, different components of the gateway were using different ways of serializing the execution request.
Some of them were ignoring `variables` if it is empty, some of not, this was causing the signature generation to be different for the same query.
For example, it was working as expected in Proxy mode, but not working as expected in Federation Gateway mode.

With this change, now we have a shared helper to serialize the upstream execution request with a memoized `print` function for query AST etc to have a consistent serialization so consistent signature generation for HMAC.


.changeset/chilled-pianos-destroy.md
```diff
@@ -0,0 +1,15 @@
...
+'@graphql-mesh/transport-http': patch
+'@graphql-tools/executor-http': patch
+'@graphql-mesh/transport-ws': patch
+---
+
+Like HMAC Upstream Signature plugin, different components of the gateway were using different ways of serializing the execution request.
+Some of them were ignoring `variables` if it is empty, some of not, this was causing the signature generation to be different for the same query.
+For example, it was working as expected in Proxy mode, but not working as expected in Federation Gateway mode.
+
+With this change, now we have a shared helper to serialize the upstream execution request with a memoized `print` function for query AST etc to have a consistent serialization so consistent signature generation for HMAC.
```
dotansimha, line 15: should we add an example for the helper? or link to the docs/example?

dotansimha (approved):

.changeset/chilled-pianos-destroy.md
ardatan: Done!

PR #387, opened 2024-12-27
fix(federation): shared mutation field should not be batched

ardatan:
In case of shared root field on Mutation, it was batched incorrectly across subgraphs. But instead only one mutation should be called as mutations should not be parallel


Urigo (approved):
Thank you @ardatan 
Is this a case that could also fit into the audit?

ardatan:
@Urigo It is actually an implementation detail(causing extra operation) not a compatibility issue.

PR #382, opened 2024-12-26
fix: batch delegation collapses two requests with different arguments into a single memoized loader

ardatan:
Originally fixed by @maxbol in the PR https://github.com/ardatan/graphql-tools/pull/5189

Closes [ #5188](https://github.com/ardatan/graphql-tools/issues/5188)
Fixes [#5189](https://github.com/ardatan/graphql-tools/pull/5189)

PR #396, opened 2024-12-31
fix(batch-delegate): memoize the key arguments correctly

ardatan:
Fixes the regression introduced in https://github.com/graphql-hive/gateway/pull/382
And breaks tests here https://github.com/ardatan/graphql-mesh/pull/8213

This pr fixes the memoization/caching the key arguments correctly;

With the following schema and resolvers, `userById` should batch all the requests to `usersByIds`;

```ts
{
      typeDefs: /* GraphQL */ `
        type User {
          id: ID!
          email: String!
        }
        type Query {
          userById(id: ID!): User
          usersByIds(ids: [ID!]): [User]
        }
      `,
      resolvers: {
        Query: {
          usersByIds: (_root, args) => {
            return args.ids.map((id: string) => users.find((user) => user.id === id));
          },
          userById: (root, args, context, info) => {
            return batchDelegateToSchema({
              schema: userSubschema,
              fieldName: 'usersByIds',
              key: args.id,
              rootValue: root,
              context,
              info,
            })
          },
        },
      },
    }
```

This query should batch all the requests to `usersByIds`:

```graphql
{
  userById(id: "1") {
    id
    email
  }
  userById(id: "2") {
    id
    email
  }
}
```

The delegation request should be;

```graphql
{
  usersByIds(ids: ["1", "2"]) {
    id
    email
  }
}
```

### DIFF
--- a/packages/batch-delegate/tests/cacheByArguments.test.ts
+++ b/packages/batch-delegate/tests/cacheByArguments.test.ts
@@ -123,26 +123,34 @@ describe('non-key arguments are taken into account when memoizing result', () =>
     const query = /* GraphQL */ `
       query {
         trendingChirps {
-          withObfuscatedEmail: chirpedAtUser(obfuscateEmail: true) { # Batch 1
+          withObfuscatedEmail: chirpedAtUser(obfuscateEmail: true) {
+            # Batch 1
             id
             email
-            friendsWithoutObfuscatedEmail: friends(obfuscateEmail: false) { # Batch 3
+            friendsWithoutObfuscatedEmail: friends(obfuscateEmail: false) {
+              # Batch 3
               id
               email
             }
-            friendsWithObfuscatedEmail: friends(obfuscateEmail: true) { # Batch 4
+            friendsWithObfuscatedEmail: friends(obfuscateEmail: true) {
+              # Batch 4
               id
               email
             }
           }
-          withoutObfuscatedEmail: chirpedAtUser(obfuscateEmail: false) { # Batch 2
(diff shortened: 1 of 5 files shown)


packages/batch-delegate/tests/cacheByArguments.test.ts
```diff
@@ -110,4 +119,66 @@ describe('non-key arguments are taken into account when memoizing result', () =>
     expect(chirps[0].withoutObfuscatedEmail.email).toBe(`1@test.com`);
     expect(chirps[1].withoutObfuscatedEmail.email).toBe(`2@test.com`);
   });
+  test('memoizes key arguments as part of batch delegation', async () => {
```
Urigo, line 122: can we expand the tests to cover more cases?

like more nested levels?
or more then two?

ardatan: Good point! Added one tests deeper levels...

Urigo: thanks!
what about if the arguments are in deeper nested levels and not on the root level?

ardatan: Done!

EmrysMyrddin (approved):

packages/batch-delegate/tests/cacheByArguments.test.ts
dotansimha: great point @Urigo. thanks for fixing @ardatan

dotansimha (approved):

PR #412, opened 2025-01-06
Dont install peer dependencies automatically with npm in Docker

enisdenjo:
Installing with `npm` includes peer dependencies by default - which is not ideal when [extending the Docker image](https://the-guild.dev/graphql/hive/docs/gateway/deployment/docker#extend-docker-image) because `graphql` is often a peer dependency and installing it again will create a duplicate module leading to possible multiple instances of `graphql` objects.

As a solution, an `.npmrc` file is included in the image containing an `omit=peer` which will install everything except for peer dependencies.

`yarn` and `pnpm` are not affected (they don't auto-install peer dependencies).

### TODO

- [x] Improve e2e to use headers as means of authenticating (improves example too)


dotansimha (approved):

e2e/operation-field-permissions/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@e2e/operation-field-permissions",
+  "private": true,
+  "dependencies": {
+    "@envelop/core": "^5.0.2",
```
ardatan, line 5: Do we always need that?

enisdenjo: We need it because:
1. the example converter copies the package.json, so we want to include all peers there for the example to work as expected
1. core is a peer dep of the plugin - without it in the package.json, yarn will raise warnings

Issue #419, opened 2025-01-08
Downstream requests aborted when refreshing schema


fauna5:
### Issue

We refresh our `supergraph` schema by setting `pollingInterval: 120_000` - this works fine. However, if there are any live downstream requests when a refresh happens then these are aborted with the following message:
```
{
  "errors": [
    {
      "message": "The operation was aborted. reason: Error: Executor was disposed.",
      "path": [
        "hello"
      ],
      "extensions": {
        "code": "DOWNSTREAM_SERVICE_ERROR",
        "request": {
          "method": "POST",
          "body": "{\"query\":\"{__typename hello}\"}"
        }
      }
    }
  ],
  "data": {
    "hello": null
  }
}
```

### Example recreation
I have created an example that shows this occurring - https://github.com/fauna5/hive-gateway-playground/tree/main . In the example, the request takes 20s and we refresh the schema every 5s. Due to the extreme configuration it happens 100% of the time. In production this happens quite frequently due to our request volumes.

### Specific configuration
It only happens when the following occur:
a) Using the `mesh` sourceHandler `loadGraphQLHTTPSubgraph`. I tried to recreate this with a REST API and didn't observe the behaviour
b) Passing headers that need to read the context
```
      operationHeaders: {
            Authorization: "{context.headers['authorization']}",
          }
```
Deleting this config fixes the behaviour.

### Environments
```
OSX
Node 22
"@graphql-hive/gateway": "1.7.7"
"@graphql-mesh/compose-cli": "1.3.4",
```
also seen on 
```
public.ecr.aws/docker/library/node:22-slim
@graphql-hive/gateway@1.6.6
@graphql-mesh/compose-cli@1.3.4
```

ardatan:
I tried to reproduce the issue by using your example in our E2E test suite. But it seems to work as expected.
https://github.com/graphql-hive/gateway/pull/420
Here is the PR. Could you check if I miss anything? Thanks 🙏

PR #420, opened 2025-01-08
fix(runtime): compare the results from factory functions correctly

ardatan:
**Fixes https://github.com/graphql-hive/gateway/issues/419**

Leave the supergraph configuration handling logic to fusion-runtime package so it can compare bare read supergraph sdl directly inside unified graph manager to decide if the supergraph has changed.
Before, it was parsed by gateway runtime and printed sdls might be different on manager side, so it was wrongly assuming that supergraph has changed whenever the supergraph fetcher has given by the user as a factory function.

**Another set of improvements;**

- In case of schema reload, throw `SCHEMA_RELOAD` error while recreating the transports and executors
- In case of shut down, throw `SHUTTING_DOWN` error while cleaning the transports and executors up

Previously, these errors are only thrown for subscriptions, it wasn't thrown in other type of operations as well.
And previously the thrown errors during these two cleanup and restart process were cryptic, now the mentioned two errors above are thrown with more clear messages

**No more timers**
Instead of `setTimeout`, on each request as we do on initial request, we compare the last fetch time and the current time to see if the supergraph should be refetched. We use a shared promise here to prevent race condition.
This will relax the usage in serverless, because some serverless envs complain about timers.
It will also relax the event loop which won't be busy if there is no request in the background.
In case of clustering, it will also relax the supergraph source, because when there is setTimeout, all members of the cluster will send requests at the same time to the source(CDN etc). Now they will make those requests lazily.


packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -1098,12 +1122,15 @@ function isDynamicUnifiedGraphSchema(
...
-      return false;
-    } catch {}
+    if (isValidPath(schema)) {
+      // local file path
+      return true;
+    }
+    if (isUrl(schema)) {
+      // remote url
+      return true;
+    }
```
enisdenjo, line 1132: URLs are valid paths too.
```suggestion
    // local file or remote url
    return isValidPath(schema);
```

enisdenjo (approved):
Aside from one small comment, LGTM.

packages/runtime/src/createGatewayRuntime.ts
ardatan: Oh good catch!

fauna5:
Hey - i pulled the snapshot version `1.7.9-alpha-4221681019dfd2fd156d1f27570765f312fb6cd8` and the operation was not aborted. This fixes the issue we have for now. Thanks so much for getting on to this so quickly 🫶 This'll get us to a production deploy!

It still does abort when there _is_ a schema change (this is much more rare, ~once/day). I guess it's a lot harder logic to allow the in-flight requests to complete on the old schema whilst validating new requests with the new schema. Keeping the old context around until all requests have completed. 

However as a client, when that request is aborted we don't know if the downstream request passed or failed. If that's a mutation we'd need to investigate via extra reads and logic to see if the mutation is successful. Whilst it's a tiny edge-case that we will probably not see, it still is there.

One way around this is recycling our k8s pods on schema update, but that defies the point of the `pollingInterval`

ardatan:
@fauna5 If you try the new alphas, the errors will be more clear in case of schema reload. With 503 status code and `SCHEMA_RELOAD` code inside the error, you will know that the gateway is reloading the schema. During the schema reload, any ongoing HTTP calls will be aborted and the execution will immediately be stopped to prevent leaks.
I think blocking the schema reload for those ongoing calls can be too much. Like you said, it is not that often, and can be solved with a retry logic.

ardatan:
@enisdenjo I did more changes, so could you review it again? I'm requesting another review.

packages/fusion-runtime/src/unifiedGraphManager.ts
```diff
@@ -426,7 +430,7 @@ export class UnifiedGraphManager<TContext> implements AsyncDisposable {
       );
     } catch (e) {
       this.opts.transportContext?.logger?.error('Failed to load Supergraph', e);
-      this.continuePolling();
+      this.lastLoadTime = Date.now();
```
enisdenjo, line 433: Maybe we might not set the load time here, having the next request try pulling the schema again?

ardatan: I considered polling here as an attempt and waiting for some time will give the source some backoff.

packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -671,11 +696,8 @@ export function createGatewayRuntime<
       }
 
       unifiedGraphFetcher = () =>
-        handleUnifiedGraphConfig(
-          // @ts-expect-error TODO: what's up with type narrowing
-          config.supergraph,
-          configContext,
-        );
+        // @ts-expect-error TODO: what's up with type narrowing
+        handleUnifiedGraphConfig(config.supergraph, configContext);
```
enisdenjo, line 700: This comment was intentionally set above config.supergraph argument originally because then it's allowing an error only on the argument (compared to now, allowing an error on the function and both arguments).

```suggestion
        handleUnifiedGraphConfig(
          // @ts-expect-error TODO: what's up with type narrowing
          config.supergraph,
          configContext,
        );

enisdenjo (approved):
Some small remarks, but LGTM.

PR #447, opened 2025-01-14
Improve GraphOS Support

ardatan:
Improve GraphOS supergraph fetching;

- Handle `minDelaySeconds` correctly, before retrying the supergraph request, wait for the `minDelaySeconds` to pass.
- Respect `maxRetries` (which is the maximum of the number of available uplink endpoints and 3) when fetching the supergraph.
- Try all possible uplinks before failing the supergraph request.


packages/runtime/src/fetchers/graphos.ts
```diff
@@ -0,0 +1,122 @@
...
+              attemptLogger.error(message);
+            },
+            INFO(message) {
+              attemptLogger.info(message);
+            },
+            WARN(message) {
+              attemptLogger.warn(message);
+            },
+          },
+        }).then(
```
EmrysMyrddin, line 82: Since we always return a promise, perhaps we can in this case use an `async` function ? It should have the same perfomance impact than a `then` call ?

ardatan: You're right! Changed it.

packages/runtime/tests/graphos.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      const { unifiedGraphFetcher } = createTestFetcher(
+        { fetch: mockFetchError },
+        { maxRetries: 3 },
+      );
+
+      const result = unifiedGraphFetcher({}).catch((err) => err);
+      for (let i = 0; i < 3; i++) {
+        await advanceTimersByTimeAsync(1_000);
+      }
+      expect(await result).toBeInstanceOf(Error);
```
EmrysMyrddin, line 58: Isn't that weird ? It should not resolve to an error ? It should throw and error ?

ardatan: I made it resolve the error in the catch clause;
https://github.com/graphql-hive/gateway/pull/447/files#diff-6f0ac42f390077837143cf368831d9825e51dbba58dcb9b4e6eb77d4cd3cd5f4R54
Because Bun tester throws the error instead of catching it with `expect`

EmrysMyrddin (approved):

PR #462, opened 2025-01-15
@graphql-hive/importer and point to exact syntax error while parsing config

enisdenjo:
Closes https://github.com/ardatan/graphql-mesh/issues/8267

`@graphql-hive/importer` is pretty much a slimmed down and improved version of [`@graphql-hive/include`](https://github.com/ardatan/graphql-mesh/tree/master/packages/include)
 - No jiti
 - Zero-dependency bundle
 - Filepaths in syntax errors
 - Type-safer
 - Tested transpiler

### Other improvements

- Processes that have exitted before test end wont exit again in the leftover stack
- Docker container will be added to the leftover stack only if the container was actually started
- No more leftover child processes after E2Es complete

### TODO

- [x] Have prettier and typescript ignore the syntax error file(s)
- [ ] https://github.com/ardatan/graphql-mesh/pull/8280
- [ ] Test the registration hook (vitest injects its own transpiler so it's not that simple, maybe later)

### DIFF
--- a/internal/proc/src/index.ts
+++ b/internal/proc/src/index.ts
@@ -116,7 +116,10 @@ export function spawn(
         await terminate(child.pid);
       }
       child.kill();
-      await waitForExit;
+      await waitForExit.catch(() => {
+        // we dont care about if abnormal exit code when disposing
+        // specifically in Windows, exit code is always 1 when killing a live process
+      });
     },
   };
   stack?.use(proc);


internal/e2e/src/tenv.ts
```diff
@@ -526,7 +526,6 @@ export function createTenv(cwd: string): Tenv {
             path.resolve(__project, 'packages', 'gateway', 'src', 'bin.ts'),
             ...getFullArgs(),
           );
-          leftoverStack.use(proc);
```
enisdenjo, line 529: Removed because the same process was added twice to the stack.

ardatan (approved):

PR #471, opened 2025-01-17
fix(delegate): spread on unions

ardatan:
While creating a delegation request for the subschema, an selection set should be spreaded on the union type field correctly.

In case of the following schema;

```graphql
type Query {
    foo: Foo
}

union Foo = Bar | Baz

type Bar {
    id: ID!
    name: String
    age: Age
}

type Age {
    years: Int
    months: Int
}

type Baz {
    id: ID!
    name: Name
    age: Int
}

type Name {
    first: String
    last: String
}
```

If the operation is generated as following;
```graphql
query {
    foo {
        id
        name
        age {
            years
            months
        }
    }
}
```

It should be spreaded on the union type field correctly as following;
```graphql
query {
    foo {
        ... on Bar {
            id
            age {
                years
                months
            }
        }
        ... on Baz {
            id
            name {
                first
                last
            }
        }
    }
}
```

### DIFF
--- a/.changeset/beige-experts-rule.md
+++ b/.changeset/beige-experts-rule.md
@@ -1,0 +1,73 @@
+---
+'@graphql-tools/delegate': patch
+---
+
+While creating a delegation request for the subschema, an selection set should be spreaded on the union type field correctly.
+
+In case of the following schema;
+
+```graphql
+type Query {
+    foo: Foo
+}
+
+union Foo = Bar | Baz
+
+type Bar {
+    id: ID!
+    name: String
+    age: Age
+}
+
+type Age {
+    years: Int
+    months: Int
+}
+
+type Baz {
+    id: ID!
+    name: Name
+    age: Int
+}
+
+type Name {
+    first: String
+    last: String
+}
+```
+
+If the operation is generated as following;
+```graphql
+query {
+    foo {
+        id
+        name
+        age {
+            years
+            months
+        }
+    }
+}
+```
+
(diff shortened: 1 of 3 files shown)


packages/delegate/src/finalizeGatewayRequest.ts
```diff
@@ -547,7 +566,7 @@ function finalizeSelectionSet(
             return null;
           }
           const namedType = getNamedType(type);
-          if (!schema.getType(namedType.name) == null) {
+          if (schema.getType(namedType.name) == null) {
```
dotansimha, line 569: is this change related? can you maybe elaborate? 
I'm asking because there might be a simpler way to flip the condition? 😆

.changeset/beige-experts-rule.md
```diff
@@ -0,0 +1,73 @@
+---
+'@graphql-tools/delegate': patch
+---
+
+While creating a delegation request for the subschema, an selection set should be spreaded on the union type field correctly.
```
dotansimha, line 5: That's a great changeset! Thank you!

dotansimha (commented):
@ardatan nice! 
Please see my comment about one of the changes. Also please note that some CI checks are failing?

packages/delegate/src/finalizeGatewayRequest.ts
ardatan: This condition was already incorrectly written. It should be like if the visited type is not found, return null.

enisdenjo (approved):

PR #481, opened 2025-01-20
Upgrade to graphql-ws v6 and release v1 of WebSocket transport

enisdenjo:
> [!IMPORTANT]
> This introduces a breaking change and will request major bump to affected packages.

Ref GW-49

### TODO
- [x] Fix WS transport tests (currently using mocking but now passing a custom client because of the extra client options)
- [x] Leaks leaks leaks...


packages/transports/ws/tests/ws.spec.ts
```diff
@@ -2,93 +2,129 @@ import type {
...
+    },
   });
 
+  const ws = await createDisposableWebSocketServer();
+
+  const onConnectFn = vi.fn<NonNullable<TServerOptions['onConnect']>>();
+
+  useServer({ schema, onConnect: onConnectFn, ...opts }, ws.server);
+
+  const executor = wsTransport.getSubgraphExecutor({
```
ardatan, line 40: This also needs to be disposed, no?

.changeset/@graphql-hive_gateway-481-dependencies.md
```diff
@@ -0,0 +1,7 @@
+---
+'@graphql-hive/gateway': patch
```
dotansimha, line 2: I think the "eventual" changeset here will be missing some details about the upgrades? 
@enisdenjo can you please add more info to the main release changeset? I know the main library is not affected a lot here, but I think we should communicate it in the changeset of the gw itself and not only the internal pkgs?

dotansimha (approved):
lgtm. @enisdenjo anything to update in docs or examples? 

Also please check my comment about the changesets, and let's wait for @ardatan 's review as well.

packages/runtime/src/getGraphQLWSOptions.ts
```diff
@@ -23,17 +23,17 @@ export function getGraphQLWSOptions<TContext extends Record<string, any>, E>(
     execute: (args) => (args as EnvelopedExecutionArgs).rootValue.execute(args),
     subscribe: (args) =>
       (args as EnvelopedExecutionArgs).rootValue.subscribe(args),
-    onSubscribe: async (ctx, msg) => {
+    onSubscribe: async (ctx, _id, payload) => {
```
ardatan, line 26: This is a breaking change for v5 users. As we discussed on Slack with @enisdenjo, we will have a fallback logic here to keep supporting v5.

packages/transports/ws/tests/ws.spec.ts
```diff
@@ -2,93 +2,165 @@ import type {
...
+    url = server.url.toString().replace('http', 'ws');
+    dispose = () => server.stop(true);
+  } else {
+    const ws = await createDisposableWebSocketServer();
+    useServer({ schema, onConnect: onConnectFn }, ws.server);
+    url = ws.url;
+    dispose = () => ws[DisposableSymbols.asyncDispose]();
+  }
+
+  const executor = wsTransport.getSubgraphExecutor(
```
ardatan, line 70: Should we dispose the executor also?

.changeset/@graphql-hive_gateway-481-dependencies.md
enisdenjo: @dotansimha have you seen the other changesets?

packages/transports/ws/tests/ws.spec.ts
enisdenjo: Done.

packages/runtime/src/getGraphQLWSOptions.ts
enisdenjo: Done.

.changeset/@graphql-hive_gateway-481-dependencies.md
dotansimha: Yeah, the reason I asked it is because the rest of the changesets are for internal packages, and I wonder if we need one specifically to the root package, since this is what users usually look for?

dotansimha: we can add `'@graphql-hive/gateway': patch` to one of the other changesets? (which are explained well! thank you!)

dotansimha (approved):

.changeset/@graphql-hive_gateway-481-dependencies.md
enisdenjo: @dotansimha ok I added some where I thought it's relevant.

ardatan (approved):

PR #506, opened 2025-01-23
Fixes code scanning warnings

ardatan:
Closes https://github.com/graphql-hive/gateway/pull/492
Closes https://github.com/graphql-hive/gateway/pull/491
Closes https://github.com/graphql-hive/gateway/pull/490
Closes https://github.com/graphql-hive/gateway/pull/489
Closes https://github.com/graphql-hive/gateway/pull/488
Closes https://github.com/graphql-hive/gateway/pull/487
Closes https://github.com/graphql-hive/gateway/pull/486

### DIFF
--- a/.changeset/mean-guests-occur.md
+++ b/.changeset/mean-guests-occur.md
@@ -1,0 +1,5 @@
+---
+'@graphql-tools/delegate': patch
+---
+
+Add `isPrototypePollutingKey` to prevent accidential prototype pollution
--- a/packages/delegate/src/isPrototypePollutingKey.ts
+++ b/packages/delegate/src/isPrototypePollutingKey.ts
@@ -1,0 +1,14 @@
+const prototypePollutingKeys = [
+  '__proto__',
+  'constructor',
+  'prototype',
+] as const;
+
+type PrototypePollutingKey = (typeof prototypePollutingKeys)[number];
+
+export function isPrototypePollutingKey(
+  key: string,
+): key is PrototypePollutingKey {
+  // @ts-expect-error - typings are incorrect
+  return prototypePollutingKeys.includes(key);
+}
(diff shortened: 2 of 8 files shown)


.changeset/mean-guests-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-tools/delegate': patch
+---
+
+Add `isPrototypePollutingKey` to prevent accidential prototype pollution
```
dotansimha, line 5: @ardatan I think we can describe here that we made an internal refactor?

packages/delegate/src/isPrototypePollutingKey.ts
```diff
@@ -0,0 +1,14 @@
...
+  'constructor',
+  'prototype',
+] as const;
+
+type PrototypePollutingKey = (typeof prototypePollutingKeys)[number];
+
+export function isPrototypePollutingKey(
+  key: string,
+): key is PrototypePollutingKey {
+  // @ts-expect-error - typings are incorrect
```
dotansimha, line 12: why incorrect here? 🤔

packages/wrap/src/transforms/WrapQuery.ts
```diff
@@ -87,11 +91,18 @@ export default class WrapQuery<TContext = Record<string, any>>
       const path = [...this.path];
       while (path.length > 1) {
         const next = path.shift()!;
+        if (isPrototypePollutingKey(next)) {
```
dotansimha, line 94: what's the case here? I mean, is it passed from a user? i mean, i don't understand how this could happen?

packages/delegate/src/isPrototypePollutingKey.ts
ardatan: `includes` expects `PrototypePollutingKey` as an argument

packages/wrap/src/transforms/WrapQuery.ts
ardatan: `path` comes from the user input in the constructor.

PR #549, opened 2025-01-28
Export `getGraphQLWSOptions` function that creates `graphql-ws` for the Hive Gateway

enisdenjo:
Ref GW-124


ardatan:
Do you think we can move this to the runtime package as graphql ws itself is zero dep?

enisdenjo:
Right, that makes sense. I'll do it.

PR #554, opened 2025-01-29
Fix polling interval option

ardatan:
No need for changeset because the CLI argument was already `polling`, and I changed it to use `pollingInterval` as an option as well to align it with the runtime option `pollingInterval`(we have never released this). 
Now that change is reverted, and the runtime value is taken from `--polling` again.
Reverts this; https://github.com/graphql-hive/gateway/commit/df203610ff9ed50adb3c3c82631ecb5324648486#diff-d6cbbbfb5c53370e686cde8de203e1c0922897847f0ce2595b178af2421197aeR259
It fixes the error when having both `--polling` and `--pollingInterval` in the commanderjs configuration

### DIFF
--- a/packages/gateway/src/cli.ts
+++ b/packages/gateway/src/cli.ts
@@ -256,7 +256,7 @@ let cli = new Command()
   )
   .addOption(
     new Option(
-      '--polling, --pollingInterval <duration>',
+      '--polling <duration>',
       `schema polling interval in human readable duration`,
     )
       .default(10_000, '10s')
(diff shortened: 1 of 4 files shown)


enisdenjo (changes requested):
Why not use the `defaultOptions.polling`? Also why not have the defaultOption a string `10s` like it was before?

The reason for that was easier readibility in the CLI and matching expected input for the argument (it expects a duration string).

ardatan:
@enisdenjo It is already `10s` in the help;
https://github.com/graphql-hive/gateway/pull/554/files#diff-d6cbbbfb5c53370e686cde8de203e1c0922897847f0ce2595b178af2421197aeR262
defaultOptions contains runtime values which is `10_000`.

enisdenjo (approved):

Issue #567, opened 2025-01-30
Allow `cache` configuration to use factory functions


dotansimha:
```ts
export type KeyValueCacheFactoryFn = (
  ctx: GatewayConfigContext,
) => KeyValueCache;
```

dotansimha:
Fixed in https://github.com/graphql-hive/gateway/pull/568, will be closed with the next release

PR #568, opened 2025-01-30
improvement(cli): allow to pass a factory function to `cache` field

dotansimha:
Closes https://github.com/graphql-hive/gateway/issues/567

### DIFF
--- a/.changeset/tame-crabs-act.md
+++ b/.changeset/tame-crabs-act.md
@@ -1,0 +1,18 @@
+---
+'@graphql-hive/gateway': minor
+---
+
+Improve `cache` configuration signature. 
+
+The `cache` configuration key now allow you to pass a custom factory function to get the cache instance: 
+
+```ts
+import { defineConfig } from '@graphql-hive/gateway'
+ 
+export const gatewayConfig = defineConfig({
+  // ...
+  cache: (ctx) => {
+    
+  } 
+})
+```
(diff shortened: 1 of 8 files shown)


packages/gateway/src/config.ts
```diff
@@ -146,8 +146,12 @@ export async function getBuiltinPluginsFromConfig(
 
 export async function getCacheInstanceFromConfig(
   config: GatewayCLIBuiltinPluginConfig,
-  ctx: Pick<GatewayConfigContext, 'logger' | 'pubsub'>,
+  ctx: GatewayConfigContext,
```
dotansimha, line 149: @ardatan @enisdenjo any reason to limit the signature here? I wasn't sure if it's needed to change or not.

.changeset/tame-crabs-act.md
```diff
@@ -0,0 +1,13 @@
...
+
+Improve `cache` configuration signature. 
+
+The `cache` configuration key now allow you to pass a custom factory function to get the cache instance: 
+
+```ts
+
+```
+
+In addition, the `context` exposed to cache functions now include the complete context of the gateway.
```
ardatan, line 13: Context might be confusing here, because there is also GraphQL Context etc

packages/gateway/src/config.ts
ardatan: In order to pass the things needed, since we have a factory function to be passed, it doesn't make much sense. But maybe we can still exclude `cache`.

.changeset/tame-crabs-act.md
dotansimha: cleaned it a bit, thanks

ardatan (approved):

PR #591, opened 2025-02-04
fix(executor-graphql-ws): respect the given `lazy` and `lazyCloseTimeout` options

ardatan:
Fixes the regression introduced in https://github.com/graphql-hive/gateway/pull/481


packages/executors/graphql-ws/src/index.ts
```diff
@@ -73,18 +73,20 @@ export function buildGraphQLWSExecutor(
         }
       : WebSocket;
 
-    graphqlWSClient = createClient({
+    const clientOptions: ClientOptions = {
```
enisdenjo, line 76: Why create the clientOptions object and then create a client with them?

ardatan: Leftover :/

PR #642, opened 2025-02-12
JSON based logger

ardatan:
Blocked by https://github.com/ardatan/graphql-mesh/pull/8375

By default
- It checks NODE_ENV=production and uses pretty if it is not production, otherwise logs in JSON format

According to `LOG_FORMAT=json` or `LOG_FORMAT=pretty`, it chooses the output format.

New JSON based logger compatible with `pino-pretty`;

With pino-pretty;
image

Without pino-pretty;
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,8 @@
     "vitest": "^3.0.1"
   },
   "resolutions": {
+    "@graphql-mesh/types": "0.103.16-alpha-20250212125252-6167fff12fe4017202b35363ff2769aafe28f98a",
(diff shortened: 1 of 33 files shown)


package.json
```diff
@@ -60,6 +60,8 @@
     "vitest": "^3.0.1"
   },
   "resolutions": {
+    "@graphql-mesh/types": "0.103.16-alpha-20250212125252-6167fff12fe4017202b35363ff2769aafe28f98a",
```
dotansimha, line 63: update this before merging?

dotansimha (approved):

package.json
ardatan: Yes!

PR #742, opened 2025-02-26
Fix incorrect error path with delegating a single to a list in subschema

### LINKED ISSUE #741: Incorrect error path in mergeDataAndErrors function when delegating subschema fields in graphql-tools/delegate
## Description

The [mergeDataAndErrors](https://github.com/graphql-hive/gateway/blob/main/packages/delegate/src/checkResultAndHandleErrors.ts#L54) function in `graphql-tools/delegate` is returning an incorrect error path when we delegate a single type to a list type in a subschema during schema stitching. The error path returned from the subschema does not match the error path the supergraph returns to the client.

The issue occurs when stitching renames fields, particularly when changing from singular to plural (e.g., 'book' to 'books') to allow querying multiple items in a single request. We are delegating a single type to a list type in a subschema because we are batching requests using [TypeMerging](https://the-guild.dev/graphql/stitching/docs/approaches/type-merging), regardless of whether one item is queried or multiple.

## Steps to Reproduce

I have created a test case demonstrating this issue: https://github.com/kathrynkodama/gateway/commit/38173bd129fedf643950a63d27bb25b9d6ac044b

1. Set up a supergraph that delegates a single type to a list type in subschema  (e.g., 'book' in supergraph to 'books' in subschema).
2. Query the supergraph for a single item (e.g., `query { book { title } }`).
3. When an error occurs in the subschema, observe the returned error path.

## Expected Behaviour

The error path should reflect the query structure as sent to the supergraph, `['book', 'title']`.

## Actual Behaviour

The error path includes an unexpected array index, `['book', 0, 'title']`.

## Suspected Cause

The issue appears to stem from the path logic in: https://github.com/graphql-hive/gateway/blob/f6b65ae8bd55e5d036bb9751d9fc70cc8df7d673/packages/delegate/src/checkResultAndHandleErrors.ts#L67-L74

`path = ['book']`
`error.path = ['books', 0, 'title']`
`error.path.slice(1) = [0, 'title]` <- leaves the array index
`newPath = ['book', 0, 'title']`

Additionally, in our environment unexpected AggregateErrors are being observed, from debugging it looks like it's due to the unpathed errors: 
https://github.com/graphql-hive/gateway/blob/main/packages/delegate/src/resolveExternalValue.ts#L207-L215

## Environment
- graphql-tools/delegate: 10.2.9 (issue persists in 10.2.13)
- Apollo GraphQL server with a stitched schema
- Node.js : 22.x

kathrynkodama:
Fixes https://github.com/graphql-hive/gateway/issues/741

Ref GW-223

This test is currently failing - the supergraph is delegating a single type to a list type in subschema (e.g., 'book' in supergraph to 'books' in subschema) and forcing an error from the subschema. In the error returned to the client, the path is incorrect. 

**Expected**: The error path should reflect the query structure as sent to the supergraph, ['book', 'title'].

**Actual**: The error path includes an unexpected array index, ['book', 0, 'title'].


packages/stitch/tests/stitchSchemasPathBug.test.ts
```diff
@@ -0,0 +1,96 @@
...
+                                schema: subschema,
+                                operation: 'query' as OperationTypeNode,
+                                fieldName: 'books',
+                                args: {
+                                    ids: [args.id],
+                                },
+                                context,
+                                info,
+                            });
+                            return test[0];
```
ardatan, line 65: I think this is why you get the wrong path. Because you transform the data on your own here.
Maybe you should use batch delegation with `batchDelegateToSchema` from `@graphql-tools/batch-delegate` or 
use transforms to handle data transformation;
https://the-guild.dev/graphql/stitching/docs/transforms/operational

Because delegateToSchema expects you have the same return type.

packages/stitch/tests/stitchSchemasPathBug.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        expect(result.errors[0].message).contains('Forced error for testing purposes');
+        // path is wrong - returning as ['book', 0, 'title'] but should be ['book', "title"]
+        expect(result.errors[0].path).toEqual(['book', "title"]);
+
+        expect(result.errors[1].message).contains('Forced error for testing purposes');
+         // path is wrong - returning as ['book', 0, 'summary'] but should be ['book', "summary"]
+        expect(result.errors[1].path).toEqual(['book', "summary"]);
+    });
+});
```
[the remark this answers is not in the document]
ardatan: This is a failing test PR.

packages/stitch/tests/stitchSchemasPathBug.test.ts
```diff
@@ -0,0 +1,96 @@
...
+            type Book {
+                id: ID!
+                title: String
+                summary: String
+            }
+            `,
+            resolvers: {
+                Query: {
+                    books: (_parent, args) => ([{
+                        id: args.id,
```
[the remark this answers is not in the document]
ardatan: This is a failing test PR.

packages/stitch/tests/stitchSchemasPathBug.test.ts
```diff
@@ -0,0 +1,96 @@
...
+                type Book {
+                    id: ID!
+                    title: String
+                    summary: String
+                }
+            `,
+            resolvers: {
+                Query: {
+                    book: {
+                        selectionSet: '{ id }',
```
ardatan, line 53: Also I think this is incorrect. `selectionSet` is used to get values from the root. it is `Query` in this case;
https://the-guild.dev/graphql/stitching/docs/approaches/schema-extensions#selectionset

kathrynkodama: True, this could be omitted. I don't need `selectionSet` at all here

packages/stitch/tests/stitchSchemasPathBug.test.ts
kathrynkodama: Thanks @ardatan for your quick reply! 

Our actual setup is a little different from what I had in the tests originally. We are stitching schemas and using `merge` to resolve the fields. I've pushed an update to this test: [`dcd6b42` (#742)](https://github.com/graphql-hive/gateway/pull/742/commits/dcd6b4286e0d7f0d1b51548388826cc0478dab9f). This is truer to what we're seeing in our setup.

ardatan:
Sorry for the late response!
I pushed a fix. And CI will release alphas soon so you can try!

enisdenjo (approved):

kathrynkodama:
Thank you @ardatan!

ardatan:
Thanks @kathrynkodama for your patience!
Until it gets released, you can start using it with the rc versions here
https://github.com/graphql-hive/gateway/pull/728#issuecomment-2682315326

ardatan:
This is available in the latest release! Thanks for the feedback!

PR #746, opened 2025-02-28
feat: Subgraph Auth via AWS Sigv4

ardatan:
Docs https://github.com/graphql-hive/console/pull/6568

Completes GW-198


dotansimha (approved):
Thanks @ardatan . LGTM, simple and clean. 

Should we add e2e test as well?

PR #766, opened 2025-03-04
[GraphOS] Respect `lastSeenId` and E2E Test for GraphOS integration

ardatan:
Ref GW-232


packages/runtime/tests/graphos.test.ts
```diff
@@ -119,6 +119,25 @@ describe('GraphOS', () => {
       await advanceTimersByTimeAsync(1_000);
       expect(await result).toBe(await result2);
     });
+
+    it('should respect even if the SDL is changed', async () => {
```
ardatan, line 123:
```suggestion
    it('should respect `lastSeenId` even if the SDL is changed', async () => {
```

dotansimha (approved):

PR #795, opened 2025-03-08
refactor: handleLoggingOption & handle `string` values for `logging`

ardatan:
- Use the same logging option handling logic, and export \`handleLoggingOption\` on runtime package
- Handle string value in `logging` like `logging: 'info'`

Ref GW-242


enisdenjo (approved):
Nice.

PR #806, opened 2025-03-10
fix: type of `contentEncoding` option

ardatan:
Ref GW-243

### DIFF
--- a/.changeset/angry-countries-act.md
+++ b/.changeset/angry-countries-act.md
@@ -1,0 +1,6 @@
+---
+'@graphql-hive/gateway': patch
+'@graphql-hive/gateway-runtime': patch
+---
+
+Fix `contentEncoding` type in `defineConfig`
--- a/packages/runtime/src/types.ts
+++ b/packages/runtime/src/types.ts
@@ -542,7 +542,7 @@ interface GatewayConfigBase<TContext extends Record<string, any>> {
    *
    * [Learn more](https://graphql-hive.com/docs/gateway/other-features/performance/compression)
    */
-  contentEncoding?: false | UseContentEncodingOpts;
+  contentEncoding?: boolean | UseContentEncodingOpts;
 
   /**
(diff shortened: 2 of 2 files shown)


ardatan:
@PlayAnyData 
When you provide `contentEncoding: true`, it enables the Content Encoding plugin from `@whatwg-node/server`, in these lines;
- https://github.com/graphql-hive/gateway/blob/main/packages/runtime/src/plugins/useContentEncoding.ts#L15
- https://github.com/graphql-hive/gateway/blob/main/packages/runtime/src/plugins/useContentEncoding.ts#L27
And that plugin handles `Content-Encoding` header coming from the client/consumer to the gateway;
https://github.com/ardatan/whatwg-node/blob/master/packages/server/src/plugins/useContentEncoding.ts

In the documentation, it says you have to put `contentEncoding: true` if you want the gateway to handle compressed requests from the client to the gateway;
https://the-guild.dev/graphql/hive/docs/gateway/other-features/performance/compression#configuration-on-gateway

Do I miss anything here?

PlayAnyData:
@ardatan Okay, I think I got it. So as soon as I activate compression between gateway and subgraphs, it is also activated between client and gateway. So I cannot opt-out of client to gateway compression. Don't know, whether someone would want to do that, but this was not clear to me.
Thanks a lot for the explanation!

ardatan:
@PlayAnyData That's currently missing you're right on that one. This is done for some internal reasons today. But let us know if you need to handle that specific case!

PlayAnyData:
Nope, just wanted to activate both :-) Thanks a lot.

ardatan:
Great! @PlayAnyData Thanks for the feedback!

EmrysMyrddin (approved):
Because you want users to explicitly provide a `true` value ?

PlayAnyData:
Just for completeness, I have created this, then :-) https://github.com/graphql-hive/console/pull/6600

PR #818, opened 2025-03-12
enhance: retry the request if the schema is reloaded during the request

ardatan:
When the schema is reloaded during a query execution, retry the request instead of throwing a reload event error

Previously, it was throwing an error `SCHEMA_RELOAD` when the schema is changed during the request.

Waiting for user's confirmation to merge!


dotansimha (approved):
just add e2e test and we are good to go 🚀

enisdenjo (approved):
LGTM!

PR #832, opened 2025-03-13
poc(otel): custom hive traces

dotansimha:
Ref GW-249

The goal here is to experiment a custom OTEL SpanProcessor that will do some adjustments to make it easier to Hive Console to process the traces. 
Regular OTEL plugin should not be affected by this, as the spanprocessor will be created in an isolated environment. 

In an ideal setup, we'll do: 
```ts
hive: {
   traces: {
       token: "...",
       targetId: "..."
    }
}
```

And this will ensure to activate OTEL plugin (@EmrysMyrddin how?) and inject `HiveTracesExporter` to exporters list.

The final impl of `hive.traces: {}` plugin should ideally create this chain to create the final exporter that will talk to Hive tracing endpoint: 

```ts
const otlpExporter = new OTLPTraceExporter({
  url: 'https://hive-tracing-endpoint',
  headers: {
    'Authorization': 'Bearer hive_traces_token_here',
  },
});

const spansProcessor = new HiveTracesSpanProcessor();
```

And Hive GW will probably also need to ensure these two are added at the end of the list, because we don't want user exporters to be affected by this? (to verify..)

### DIFF
--- a/e2e/opentelemetry/gateway.config.ts
+++ b/e2e/opentelemetry/gateway.config.ts
@@ -5,6 +5,13 @@ import {
   GatewayPlugin,
 } from '@graphql-hive/gateway';
 import type { MeshFetchRequestInit } from '@graphql-mesh/types';
+import {
+  BatchSpanProcessor,
+  ReadableSpan,
+  Span,
+  SpanExporter,
+  SpanProcessor,
+} from '@opentelemetry/sdk-trace-base';
 
 // The following plugin is used to trace the fetch calls made by Mesh.
 const useOnFetchTracer = (): GatewayPlugin => {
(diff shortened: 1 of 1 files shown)


e2e/opentelemetry/gateway.config.ts
```diff
@@ -26,9 +33,142 @@ const useOnFetchTracer = (): GatewayPlugin => {
...
+                Array.from(subgraphNamesForTrace).join(','),
+              );
+            }
+          }
+        }
+      }
+    }
+
+    // For any subgraph span that's ending, make sure we capture its name
+    if (span.name && span.name.startsWith('subgraph.execute')) {
```
dotansimha, line 119: super bad impl, we should improve and see if we can depend on the values in the gw OTEL plugin, and avoid using raw strings. also make sure we don't break it by adding tests 😆

dotansimha:
just a POC, closing for now.

PR #871, opened 2025-03-17
Introduce built-in armor security features

enisdenjo:
Ref GW-227, documentation https://github.com/graphql-hive/console/pull/6630

The following built-in features are introduced:

- `maxTokens`: Limit the number of tokens in a GraphQL document.
  Defaults to `1000` tokens.
- `maxDepth`: Limit the depth of a GraphQL document.
  Defaults to `6` levels.
- `blockFieldSuggestions`: Prevent returning field suggestions and leaking your schema to unauthorized actors.
  Defaults to `true`

They are all **disabled** by default. Users have an option to enable or configure them individually. Only basic configuration options are allowed - security features needing more configuration should instead have the plugin installed and used manually.

Additionally, the `rateLimiting` option has been fixed to match whats advertised in the documentation.

#### Disable all (default)

```ts

export const gatewayConfig = defineConfig();
```

#### Enable all

```ts

  maxTokens: true,
  maxDepth: true,
  blockFieldSuggestions: true,
});
```

#### Configure

```ts

  maxTokens: 3000,
  maxDepth: 10,
  blockFieldSuggestions: false,
});
```

#### Advanced configuration

```sh
npm i @escape.tech/graphql-armor-max-depth
```

```ts
import { maxDepthRule } from '@escape.tech/graphql-armor-max-depth';

  // disable maxDepth feature by omitting it or setting it to false
  plugins: () => [
    maxDepthRule({
      ignoreIntrospection: false,
      flattenFragments: true,
    }),
  ],
});

```


ardatan (commented):
"security" name there can be confusing because we have other security features as seperate parameters on the same level.

enisdenjo:
@ardatan flattening the security option prevents the easy enabling with `security: true`.

We can move the rest of security options in side the `security` property and deprecated them. What do you think?

ardatan:
Some of other secutiry features require configuration, and `true` might not work there for them. Since this is part of Armor, armor plugins can be (maybe) grouped in `armor` instead of `security`. But grouping all security params into `security: boolean` will probably not work.

enisdenjo:
> Some of other secutiry features require configuration, and true might not work there for them

We can have sane defaults. If we cant - we just wont enable them unless configured.

Renaming to `armor` is also an option, but I am not a fan because the property itself is not self-explanatory - what is armor (for people that dont know graphql-armor)? `security` everyone knows as soon as they read it.

ardatan:
But these are not only `security` features we have. We have more. And some of other security features can't be enabled with a `true` or any defaults. It sometimes needs configuration from the user.

ardatan (approved):

PR #905, opened 2025-03-21
Testing subgraph errors in case of nullable and non-nullable field and the inaccesible subgraph

ardatan:
Ref GW-254


packages/runtime/tests/subgraph-down.test.ts
```diff
@@ -0,0 +1,212 @@
...
+            }
+            subgraph2 {
+              subgraph2Field
+            }
+          }
+        `,
+      }),
+    });
+    const res = await resp.json();
+    expect(res).toEqual({
```
dotansimha, line 102: @ardatan I would expect this to be a partial response, where `errors` is also set?

PR #922, opened 2025-03-24
Ponyfill AbortSignal.any without leaks

### LINKED ISSUE #904: Memory Leak utilizing latest gateway and node 22.14.0
Hey all,

first of all, I am not sure, if this really relates to this repo or would be better suited for @whatwg-node or a downstream repository. So, your help would be appreciated.

# Environment
We are utilizing the following versions:
latest graphql-hive/gateway
node 22.14.0 (also reproducible with 20.19.0)
reproducible in Windows/Docker

# Bug
We are facing the following issue with our gateway: With every request (it seems) the FinalizationRegistry is expanding.
Initially, the memory consumption look like this:
Image
After about 20m (of the same requests with k6), it looks like this:
Image
Then, the FinalizationRegistry contains:
Image

This keeps on growing and is not tidied up, ultimately leading to longer response times, if the gateway reaches the configured memory consumption limits.

I was thinking, maybe the node-fetch wrapper with the AbortController usage had an influence on this. However, I actually doubt it, as I also have tested a downgrade 0.7.7 where the native AbortController was not yet utilized.

I am happy to provide more information if required and for any insights what I could do to prevent this. Also happy to file bug reports somewhere else, if required.

# Reproduction
A reproduction repository is available here: https://github.com/PlayAnyData/mesh-error-duplicate/tree/reproduction/memory-leak (check the readme over there)

# Possibly related to
Could be related to: https://github.com/nodejs/node/issues/55328 - However, this has been fixed and the reproduction samples there, I could not use to reconstruct this.

enisdenjo:
Ref GW-140, ref GW-264
Closes #904


dotansimha (approved):
lgtm. let's wait also for Arda to review. Any idea why the builtin leaks in that case?

enisdenjo:
> Any idea why the builtin leaks in that case?

Not really sure, there's been some resolved issues in Node - maybe not good enough? Maybe [this](https://github.com/nodejs/node/issues/54614) is related?

Honeslty, didn't want to waste much time figuring it out. We can circle back at some time, but for now - this works.

enisdenjo:
Merging this with the http-callbacks leak in place. It requires a bigger refactor and I want to avoid pushing too much in a single release.

PR #929, opened 2025-03-25
fix(executor-http): handle empty response with HTTP error

ardatan:
Fixes https://github.com/graphql-hive/gateway/pull/177#issuecomment-2751521475


dotansimha (approved):
I think CI needs a fix? but otherwise, LGTM.

PR #933, opened 2025-03-25
Introduce Hive Gateway PubSub with hardened memory safety

enisdenjo:
Ref GW-275

### Hive PubSub

New Hive PubSub with hardened memory safety (with tests) and improved type support. Hive's PubSub is backwards compatible with Mesh PubSub so no breaking changes have been introduced. However, there _are_ some improvements like disposal that are currently not enforced, but will be in the next major.

```ts
import { PubSub } from '@graphql-hive/pubsub';

//                                      👇 topic data map 👇
const pubsub = new PubSub<{ hello: 'world'; metadata: Record<string, any> }>();

// ✅
pubsub.publish('hello', 'world');
publis.subscribe('hello', (data) => {
  // data === 'world'
});

// ✅
pubsub.publish('metadata', {});
publis.subscribe('metadata', (data) => {
  // data === Record<string, any>
});

// ❌
pubsub.publish('alloha', 'world');
publis.subscribe('alloha', () => {});

// ❌
pubsub.publish('metadata', 7);

// ✨new dispose method, will unsubscribe everyone and end all async iterables
pubsub.dispose();
pubsub[Symbol.dispose](); // supports `using pubsub = new PubSub()`
```

### Memtest flags

Introduce an easy way of changing the memtests for sped up debugging. Available flags are:

- `short` Runs the loadtest for `30s` and the calmdown for `10s` instead of the defaults.
- `heapsnaps` Takes heap snapshots instead of the defaults.
- `moreruns` Does `5` runs instead of the defaults.
- `chart` Writes the memory consumption chart.
- `sampling` Will write the heap allocation sampling profile regardless of whether the test fails.

```sh
MEMTEST=chart,short,heapsnaps,sampling yarn test:mem federation-example
```

### Other memtest stuff

Refactored and analysed heap snapshots on all active memtests - no leaks. Also increased strictness of allowed heavy frames in heap allocation frames.


ardatan (approved):

dotansimha (approved):
rebase and check ci and then we can merge, thanks!

dotansimha:
@enisdenjo can you also add a Ref link?

PR #946, opened 2025-03-26
Improvements on GraphOS and Fastify integration

ardatan:


packages/logger-pino/src/index.ts
```diff
@@ -0,0 +1,125 @@
...
+    .flatMap((messageArg) => {
+      if (typeof messageArg === 'function') {
+        messageArg = messageArg();
+      }
+      if (messageArg?.toJSON) {
+        messageArg = messageArg.toJSON();
+      }
+      if (messageArg instanceof AggregateError) {
+        return messageArg.errors;
+      }
```
EmrysMyrddin, line 24: Why do we want to flatten the list of errors ? Aren't we loosing the message of the AggregatedError ?

ardatan: We lose it but it is usually a concatenated version of child errors or a message like `Failed with multiple errors` or there might be no message as the message property of AggregateError is optional.
And keeping this kind of messages there is mostly more confusing than being helpful.

packages/runtime/src/plugins/useRetryOnSchemaReload.ts
```diff
@@ -31,13 +35,21 @@ export function useRetryOnSchemaReload<
...
         result?.errors?.some((e) => e.extensions?.['code'] === 'SCHEMA_RELOAD')
       ) {
+        let requestLogger = logger;
+        const requestId = requestIdByRequest.get(request);
+        if (requestId) {
+          requestLogger = logger.child({ requestId });
+        }
+        requestLogger.info(
+          'The operation has been aborted after the supergraph schema reloaded, retrying the operation...',
+        );
```
EmrysMyrddin, line 52: I suppose that this a temporary workaround waiting for the new Logging system ?

EmrysMyrddin (approved):

PR #965, opened 2025-03-27
feat(opentelemetry): Allow to use the plugin in Yoga

EmrysMyrddin:
This PR removes the necessity of passing the HiveGateway logger. This way, the plugin is now usable in Yoga too.

The goal is to improve our subgraphs traces while maintaining only one version of the plugin.

I didn't push the thing up to Envelop only usage, but that's something that should be doable with a bit of effort if we want to.

Related to GW-94


dotansimha (approved):
lgtm. @ardatan thoughts?

PR #990, opened 2025-04-03
chore: update prometheus plugin definition

PlayAnyData:
Updates the definition of prometheus plugin to v1 and removes todos


ardatan:
This plugin is still used by Mesh v0, that's why the implementation has those parts for backwards compatibility.

PlayAnyData:
Oh okay, that is interesting. Then, I will just drop it.

PR #993, opened 2025-04-04
feat(opentelemetry): add cache events and attributes

EmrysMyrddin:
I didn't found any standard cache related attributes, so I've add custom `gateway.cache` prefix.

I also noticed that the operation name is not set when the result is cached (since the execute hook is not called). I don't want to fix this in this PR because it will involve changes in all response cache plugin variants

Related to GW-168

### DIFF
--- a/packages/plugins/opentelemetry/src/plugin-utils.ts
+++ b/packages/plugins/opentelemetry/src/plugin-utils.ts
@@ -56,19 +56,23 @@ export function withState<
   function addStateGetters(src: any) {
     const result: any = {};
     for (const [hookName, hook] of Object.entries(src) as any) {
-      result[hookName] =
-        typeof hook !== 'function'
-          ? hook
-          : (payload: any, ...args: any[]) =>
-              hook(
-                {
-                  ...payload,
-                  get state() {
-                    return getState(payload);
-                  },
+      if (typeof hook !== 'function') {
+        result[hookName] = hook;
+      } else {
+        result[hookName] = {
+          [hook.name](payload: any, ...args: any[]) {
+            return hook(
+              {
+                ...payload,
+                get state() {
+                  return getState(payload);
                 },
-                ...args,
-              );
+              },
+              ...args,
+            );
+          },
+        }[hook.name];
+      }
     }
     return result;
   }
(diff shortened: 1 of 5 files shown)


packages/plugins/opentelemetry/src/plugin-utils.ts
```diff
@@ -78,6 +82,8 @@ export function withState<
   const pluginWithState = addStateGetters(hooks);
   pluginWithState.instrumentation = addStateGetters(instrumentation);
 
+  console.log('plugin hooks: ', Object.entries(pluginWithState));
```
dotansimha, line 85: oops

packages/plugins/opentelemetry/src/spans.ts
```diff
@@ -385,6 +395,46 @@ export function setUpstreamFetchResponseAttributes(input: {
...
+  });
+}
+
+export function recordCacheError(
+  action: 'read' | 'write',
+  error: Error,
+  payload: OnCacheGetHookEventPayload,
+) {
+  trace.getActiveSpan()?.addEvent('gateway.cache.error', {
+    'gateway.cache.key': payload.key,
```
dotansimha, line 414: good naming! 

nit: can also be `graphql.cache.key` to make it more unique in a full architecture,

dotansimha: or `graphql.response_cache.key`

dotansimha (approved):
lgtm. let's also wait for @ardatan to confirm.

e2e/opentelemetry/opentelemetry.e2e.ts
```diff
@@ -64,19 +67,56 @@ describe('OpenTelemetry', () => {
...
+          res: JaegerTracesApiResponse,
+          abort: AbortController,
+        ) => void | PromiseLike<void>,
       ): Promise<void> {
         const url = `http://0.0.0.0:${jaeger.additionalPorts[16686]}/api/traces?service=${service}`;
 
         let res!: JaegerTracesApiResponse;
         let err: any;
-        const signal = AbortSignal.timeout(15_000);
+        const timeout = AbortSignal.timeout(15_00);
```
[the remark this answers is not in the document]
ardatan: @EmrysMyrddin Should this be fixed?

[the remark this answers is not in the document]
EmrysMyrddin: It should be, but in the v2 branch directly. I will do it on next rebase.

enisdenjo:
Heads up, I've rebased v2 on main. There's conflicts now.

ardatan (approved):

PR #997, opened 2025-04-04
Configure Subgraph Error handling

ardatan:
Ref GW-291

- **BREAKING**: HTTP Executor no longer takes `serviceName` as an option.
- Both HTTP executor and `@graphql-mesh/transport-http-callback` no longer handle `DOWNSTREAM_SERVICE_ERROR` error code with `serviceName`.
- Gateway runtime handles subgraph errors on its own with `DOWNSTREAM_SERVICE_ERROR` error code and `serviceName` as a property. This behavior can be configured with `subgraphErrors` option of the `createGatewayRuntime` function or CLI config.

```ts
subgraphError: {
   errorCode: 'DOWNSTREAM_SERVICE_ERROR', // or `false` to remove this code completely
   subgraphNameProp: 'serviceName' // or `false` to remove this prop completely
}
```


.changeset/quick-buses-obey.md
```diff
@@ -0,0 +1,17 @@
+---
+'@graphql-tools/executor-http': major
+'@graphql-mesh/transport-http-callback': minor
+'@graphql-hive/gateway-runtime': minor
+'@graphql-tools/federation': patch
+---
+
+- **BREAKING**: HTTP Executor no longer takes `serviceName` as an option.
```
dotansimha, line 8: Any reason for the breaking change? or is it just internal and GW is not affected?

dotansimha (approved):
lgtm. let's wait for Laurin to confirm as well. Also, please check my comment about the breaking change.

.changeset/quick-buses-obey.md
ardatan: No breaking changes for the GW

packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -991,6 +992,16 @@ export function createGatewayRuntime<
...
+  if (config.subgraphErrors !== false) {
+    basePlugins.push(
+      useSubgraphErrorPlugin(
+        typeof config.subgraphErrors === 'object'
+          ? config.subgraphErrors
+          : undefined,
+      ),
+    );
+  }
+
```
enisdenjo, line 1004: Isn't this breaking? You now have to explicitly enable `subgraphErrors` in the config.

ardatan: Yes it is enabled by default already. I just make it configurable with this PR.

enisdenjo: Understood.

enisdenjo (approved):

PR #1030, opened 2025-04-09
Hive Logger

enisdenjo:
Ref GW-201, ref GW-205, ref GW-204, ref GW-279, ref GW-172

### TODO

- [ ] Use logger from context in OpenTelemetry
- [ ] Deprecate logger-json
- [ ] Deprecate logger-pino
- [ ] Deprecate logger-winston


packages/logger/src/writers/pino.ts
```diff
@@ -0,0 +1,18 @@
...
+  #pinoLogger: PinoLogger;
+  constructor(pinoLogger: PinoLogger) {
+    this.#pinoLogger = pinoLogger;
+  }
+  write(
+    level: LogLevel,
+    attrs: Attributes | null | undefined,
+    msg: string | null | undefined,
+  ): void {
+    this.#pinoLogger[level](attrs, msg || undefined);
```
ardatan, line 16: How does this integrate with Pino's child logic?

enisdenjo: Nope, all of the metadata and child mechanism is handled by Hive Logger (since you're already using it) and Pino and Winston (and all other writers) simply write.

enisdenjo: All of their writing mechanisms (like Pino and Winstons transports) will still work as expected.

packages/logger/src/writers/console.ts
```diff
@@ -0,0 +1,124 @@
...
+  noColor?: boolean;
+  /**
+   * Whether to include the timestamp at the beginning of the log message.
+   *
+   * @default false
+   */
+  noTimestamp?: boolean;
+}
+
+export class ConsoleLogWriter implements LogWriter {
```
ardatan, line 34: Do we still respect custom inspection `Symbol.for('nodejs.util.inspect.custom');` or from now on we just use `JSON.stringify`?

ardatan: Nvm found it! Great!

enisdenjo: [We do](https://github.com/graphql-hive/gateway/blob/a39189a656671ed9ab571bd39edba5c94524410d/packages/logger/src/utils.ts#L131-L162), object inspection is now implemented in the logger itself and not the writers ([see `parseAttrs`](https://github.com/graphql-hive/gateway/blob/a39189a656671ed9ab571bd39edba5c94524410d/packages/logger/src/utils.ts#L57C17-L57C27)).

packages/logger/tests/writers.test.ts
```diff
@@ -0,0 +1,74 @@
+import { describe, expect, it } from 'vitest';
+import { Logger } from '../src/logger';
+import { ConsoleLogWriter, ConsoleLogWriterOptions } from '../src/writers';
+import { stableError } from './utils';
+
+describe.skipIf(
+  // bun is serialising the snapshots differently. the object keys are out of order...
```
ardatan, line 7: `fast-json-stringify` doesn't work with this?

enisdenjo: It does, it's just that bun serialises the snapshots differently putting the object keys in different order than node.

packages/fusion-runtime/src/executor.ts
```diff
@@ -70,9 +70,7 @@ export function getExecutorForUnifiedGraph<TContext>(
     enumerable: true,
     get() {
       return function unifiedGraphExecutorDispose() {
-        opts?.transportContext?.logger?.debug(
-          'Disposing unified graph executor',
-        );
+        opts?.transportContext.log.debug('Disposing unified graph executor');
```
ardatan, line 73: `transportContext` can be different in programmatic usage, so it might not have the same context with the gateway because this is just the runtime

packages/transports/http-callback/src/index.ts
```diff
@@ -138,8 +138,10 @@ export default {
           stopSubscription(createTimeoutError());
         }, heartbeatIntervalMs),
       );
-      subscriptionLogger?.debug(
-        `Subscribing to ${transportEntry.location} with callbackUrl: ${callbackUrl}`,
+      log.debug(
+        'Subscribing to %s with callbackUrl: %s',
```
ardatan, line 142: What is the difference between this and template literals?

packages/fusion-runtime/src/executor.ts
enisdenjo: I made sure that the `transportContext` always has a log, updated the typings too.

packages/logger/tests/writers.test.ts
ardatan: I actually meant can we use that library to serialize the objects then take the snapshot to make it consistent between test runners?

ardatan: https://github.com/graphql-hive/gateway/pull/1030/files/a39189a656671ed9ab571bd39edba5c94524410d#r2103070734

e2e/hmac-auth-https/services/users/index.ts
```diff
@@ -20,6 +21,7 @@ const yoga = createYoga({
   logging: true,
   plugins: [
     useHmacSignatureValidation({
+      log: new Logger({ level: 'debug' }),
```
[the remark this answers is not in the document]
ardatan: I know noone really does but just wanted to make sure the name is good enough. But if it is a standard, we can follow too 👍

packages/logger/src/writers/pino.ts
ardatan: If it really doesn't matter to use their metadata and child features, then it is fine. I was concerning about if they have a special logic for their own child and metadata stuff.

packages/transports/http-callback/src/index.ts
enisdenjo: I've [added benchmarks](https://github.com/graphql-hive/gateway/pull/1030/commits/2c512a9004f7e2ac471bde15cf9dd9b064d6f87e), the overhead is minimal (template literals being on average 1.10x faster).

However, using the interpolation approach has 2 benefits:
- You have useful placeholders like `%o` for formatting objects right there
- It's easier to read because the placeholders are shorter therefore having the whole log message right there

Nevertheless, it's just a convention you'll see out there. We can switch back to using the literals in the gateway if you want (even though we'll keep the interpolation feature).

packages/logger/tests/writers.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    hello
+    world
+  }
+}
+`,
+      },
+      'graphql',
+    );
+
+    expect(logs).toMatchSnapshot();
```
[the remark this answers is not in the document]
enisdenjo: `err` properties were still out of order. But I just dropped the error because there's no specific colors applied to it. [Now it works](https://github.com/graphql-hive/gateway/pull/1030/commits/7c4daa14687333c83ca23d8c74c7779e1bc5ac03), thanks!

packages/transports/http-callback/src/index.ts
ardatan: I am just afraid the overhead/latency might grow silently (exponentially also) with "safe" stringifying and formatting phases. On the other hand, we don't usually print obejcts that often especially by default

packages/fusion-runtime/src/executor.ts
enisdenjo: Ah, have overseen that one. I'll make a todo, thanks for pointing it out!

enisdenjo: Or maybe just keep the transportContext option? 🤔

ardatan: Or we can have build a default transport context but this might be an extra for this PR.

PR #1096, opened 2025-05-08
Add response log under tmp/codegen when JSON parsing failed

amkisko:
Hey!

Here's suggestion on improving logging experience for error, server response might have HTML response or anything else what is not possible to read in console and requires accessing with editor.

I would like to participate in the discussion how to make it better, do we need a configuration variable here for debugging purposes only so that it does not create files without conditionals.

Thank you!

### DIFF
--- a/packages/executors/http/src/index.ts
+++ b/packages/executors/http/src/index.ts
@@ -30,6 +30,7 @@ import {
   createResultForAbort,
   hashSHA256,
 } from './utils.js';
+import * as fs from 'fs';
 
 export type SyncFetchFn = (
   url: string,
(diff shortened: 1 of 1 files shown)


enisdenjo:
Currently there work being done in #1030 which will significantly improve the logging experience in the gateway. Your opinion there would be helpful instead.

What you did here does not really work for us, the gateway's transports are expected to run in all JS enviornments - including those that dont have access to the filesystem - like Cloudlflare Workers and other edge lambdas; meaning, writing files is a no-go.

Instead, you can write a plugin using the `onFetch` hook returning a function which is an `onFetchDone` hooks to inspect the result, something roughly like this:
```ts
import fs from 'node:fs/promises';

  plugins: () => [
    {
      onFetch() {
        return async ({ response }) => {
          if (!response.ok) {
            // write the response body to a file when response is failing
            const content = await response.text();
            await fs.writeFile('./somewhere.txt', content);
          }
        };
      },
    },
  ],
});

```

amkisko:
@enisdenjo thanks for detailed explanation, I was not aware about `onFetch` and have not seen its usage in the code much, need to assess it, seems to be exact thing I need for temporary hook.

amkisko:
I think the main problem will be to somehow patch gateway usage in graphql-codegen library in order to get closer to this. :)

enisdenjo:
What do you mean by that? How do you use graphql-codegen with the gateway, do you have reproduction we can look at?

enisdenjo:
Ah ok, so you're actually talking about the http executor in graphql-codegen, not the gateway. I got confused for a moment.

amkisko:
Yes, that's correct.

PR #1113, opened 2025-05-14
Subgraph serving fixes

ardatan:
Handle serving subgraphs when;
- [x] No entity is found
- [x] Query root type has a different name than `Query`
- [x] Entity is defined with federation transform

Also - Do not print warnings for missing `gatewayConfig` export for `mesh.config` files

### DIFF
--- a/.changeset/slimy-donuts-press.md
+++ b/.changeset/slimy-donuts-press.md
@@ -1,0 +1,5 @@
+---
+'@graphql-hive/gateway': patch
+---
+
+Do not print warnings for missing `gatewayConfig` export for `mesh.config` files


e2e/openapi-subgraph/openapi-subgraph.e2e.ts
```diff
@@ -37,13 +37,14 @@ describe('openapi-subgraph', () => {
       replaceDockerHostNamesBack(result),
     );
   });
-  it('resolves entitites correctly', async () => {
+  it.only('resolves entitites correctly', async () => {
```
enisdenjo, line 40:
```suggestion
  it('resolves entitites correctly', async () => {
```

e2e/openapi-subgraph/openapi-subgraph.e2e.ts
```diff
@@ -37,13 +37,14 @@ describe('openapi-subgraph', () => {
...
     const { execute } = await gateway({
       subgraph: {
         with: 'mesh',
         services: [OASService, GQLService],
         subgraphName: 'OAS',
       },
+      pipeLogs: true,
```
enisdenjo, line 47:
```suggestion
```

enisdenjo (approved):

PR #1117, opened 2025-05-15
Optimize query planner for `@provides`

ardatan:
Ref GW-312

Optimizes `@provides` handling by avoiding the generation of new query plans when a parent subgraph already supplies the requested fields.  
- Refactors and inlines `subtractSelectionSets` to compute leftover selections.  
- Threads a `providedSelectionNode` through planning to subtract out provided fields early.  
- Updates stitching and federation logic to conditionally skip planning when selections are already available.

Note that, this is not a fix for federation compatibility, it is an optimization for the query planner.

### DIFF
--- a/e2e/type-merging-batching/type-merging-batching.e2e.ts
+++ b/e2e/type-merging-batching/type-merging-batching.e2e.ts
@@ -33,63 +33,63 @@ it.each([
       }
     `,
   },
-  {
-    name: 'Authors',
-    query: /* GraphQL */ `
-      query Authors {
-        authors {
-          id
-          name
-          books {
-            id
-            title
-            author {
-              id
-              name
-            }
-          }
-        }
-      }
-    `,
-  },
-  {
-    name: 'Book',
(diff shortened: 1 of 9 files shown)


e2e/type-merging-batching/type-merging-batching.e2e.ts
```diff
@@ -33,63 +33,63 @@ it.each([
...
+  //           name
+  //           books {
+  //             id
+  //             title
+  //           }
+  //         }
+  //       }
+  //     }
+  //   `,
+  // },
```
enisdenjo, line 92: Should this stay commented out?

enisdenjo (changes requested):

e2e/type-merging-batching/type-merging-batching.e2e.ts
ardatan: Oh good catch. I left it by mistake.

enisdenjo (approved):

PR #1128, opened 2025-05-16
Update Docker images -> Node 24 and Bun Slim

ardatan:


enisdenjo (approved):
Lets merge when CI passes, great!

PR #1150, opened 2025-05-23
feat(wrap): allow making the new field type Nullable

MkMan:
## Description

Allow providing an optional configuration object to control the nullability of the new Wrapper field.

## Why

While recently investigating an incident we suffered I discovered that fields Wrapped under a new field are automatically marked as required (NonNull).

In our case this caused a problem because the fields were from an external stitched schema and when the API returned `null` our service crashed as it expected something back.

## Comments

I couldn't find contributing guidelines so please provide me any feedback needed to make this change in a way that aligns with the repo's standards and direction.

Many thanks 😃

### DIFF
--- a/.changeset/brave-yaks-camp.md
+++ b/.changeset/brave-yaks-camp.md
@@ -1,0 +1,5 @@
+---
+'@graphql-tools/wrap': minor
+---
+
+Allow making the new wrapper field type Nullable


enisdenjo (changes requested):
This is looking good! Just please create a changeset using `yarn changeset` to get this ready for release. Thanks!

MkMan:
> This is looking good! Just please create a changeset using `yarn changeset` to get this ready for release. Thanks!

Done, thanks for reviewing it @enisdenjo 🙂

Issue #1149, opened 2025-05-22
Add support for OpenTelemetry resource attributes configuration


Dzheky:
Currently, Hive Gateway's OpenTelemetry configuration only supports `serviceName` but doesn't support standard resource attributes like `deployment.environment`, `service.version`, etc.

Request: Add support for either:
1. Standard OpenTelemetry environment variables (OTEL_RESOURCE_ATTRIBUTES, OTEL_SERVICE_NAME, etc.)
2. A `resource` configuration option in the openTelemetry config

PR #1151, opened 2025-05-23
feat(opentelemetry): add resource configuration support

Dzheky:
## Description
Adds support for custom resource attributes configuration in the OpenTelemetry plugin, enabling users to specify additional resource metadata like `deployment.environment`, `service.instance.id`, and other custom attributes.

## Example
```typescript
{
  //...
  openTelemetry: {
    exporters: [createOtlpHttpExporter({ url: 'http://localhost:4318' })],
    serviceName: 'my-gateway',
    resource: {
      'deployment.environment': 'production',
      'service.instance.id': process.env.HOSTNAME,
      'service.namespace': 'platform',
      'service.name': 'custom-service-name', // Overrides serviceName
    },
  }
  //...
}
```

## Why
Previously, users could only configure service.name via the serviceName option. This limitation made it difficult to add essential resource attributes like deployment environment, instance IDs, or custom metadata that are needed for proper observability and trace correlation in production environments.

Related Issues
#1149

### DIFF
--- a/packages/plugins/opentelemetry/src/plugin.ts
+++ b/packages/plugins/opentelemetry/src/plugin.ts
@@ -23,7 +23,7 @@ import {
   type Tracer,
 } from '@opentelemetry/api';
 import { setGlobalErrorHandler } from '@opentelemetry/core';
-import { Resource } from '@opentelemetry/resources';
+import { Resource, type ResourceAttributes } from '@opentelemetry/resources';
 import { type SpanProcessor } from '@opentelemetry/sdk-trace-base';
 import { WebTracerProvider } from '@opentelemetry/sdk-trace-web';
 import { handleMaybePromise } from '@whatwg-node/promise-helpers';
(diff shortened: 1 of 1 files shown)


enisdenjo:
@EmrysMyrddin care to take a look? Is this something we'll have in v2?

EmrysMyrddin:
Yes, this is something that will be possible in v2. I'm currently working on configuration changes!

Anyway, it is already possible, you can disable automatic sdk initialization and do your own OTEL setup. I would recommend this over the automatic configuration for a production configuration, it will give you the full control of OTEL behavior.

Dzheky:
@EmrysMyrddin This is how we currently did this, it would be still nice to have this in automatic initialisation. So should this PR be closed and it will be implemented in v2?

EmrysMyrddin:
@Dzheky The new OTEL configuration PR has landed here: https://github.com/graphql-hive/gateway/pull/1171

You can find the documentation here : https://pr-6852.hive-landing-page.pages.dev/docs/gateway/monitoring-tracing#opentelemetry-traces

We would be happy to have your feedbacks on this new configuration approach :-)

PR #1171, opened 2025-05-26
feat(opentelemetry): correct span parenting + fix batching support

EmrysMyrddin:


internal/e2e/src/tenv.ts
```diff
@@ -423,7 +423,7 @@ export function createTenv(cwd: string): Tenv {
         case 'bun-docker':
         case 'docker': {
           const volumes: ContainerOptions['volumes'] =
-            runner?.docker?.volumes || [];
+            runner?.docker?.volumes?.concat() || [];
```
enisdenjo: concat what? 
```suggestion
            runner?.docker?.volumes || [];
````

File: e2e/opentelemetry/package.json [hunk 2]
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@e2e/opentelemetry",
+  "version": "1.0.0",
```
enisdenjo:
```suggestion
```

internal/testing/src/getLocalhost.ts
```diff
@@ -12,14 +12,11 @@ export async function getLocalhost(
     for (const hostname of hostnames) {
       const url = `${protocol}://${hostname}:${port}`;
       if (isDebug()) {
-        console.log(`getLocalhost(port): Trying ${url}`);
+        console.log(`getLocalhost(port): Trying ${protocol} on port ${port}`);
```
enisdenjo: Why not full url? One benefit of having the full url is that you can click on the link from the terminal and open in browser.

package.json
```diff
@@ -66,10 +66,10 @@
...
     "@graphql-tools/delegate": "workspace:^",
-    "@opentelemetry/exporter-trace-otlp-http@npm:0.200.0": "patch:@opentelemetry/exporter-trace-otlp-http@npm%3A0.200.0#~/.yarn/patches/@opentelemetry-otlp-exporter-base-npm-0.200.0-6fb25211c7.patch",
-    "@opentelemetry/otlp-exporter-base@npm:0.200.0": "patch:@opentelemetry/otlp-exporter-base@npm%3A0.200.0#~/.yarn/patches/@opentelemetry-otlp-exporter-base-npm-0.200.0-6fb25211c7.patch",
-    "@opentelemetry/resources@npm:2.0.0": "patch:@opentelemetry/resources@npm%3A2.0.0#~/.yarn/patches/@opentelemetry-resources-npm-2.0.0-f20376a5f9.patch",
+    "@opentelemetry/otlp-exporter-base@npm:0.202.0": "patch:@opentelemetry/otlp-exporter-base@npm%3A0.202.0#~/.yarn/patches/@opentelemetry-otlp-exporter-base-npm-0.202.0-f6f29c2eeb.patch",
+    "@rollup/plugin-node-resolve@npm:^15.2.3": "patch:@rollup/plugin-node-resolve@npm%3A16.0.1#~/.yarn/patches/@rollup-plugin-node-resolve-npm-16.0.1-2936474bab.patch",
     "@vitest/snapshot": "patch:@vitest/snapshot@npm:3.1.2#~/.yarn/patches/@vitest-snapshot-npm-3.1.1-4d18cf86dc.patch",
+    "ansi-color@npm:^0.2.1": "patch:ansi-color@npm%3A0.2.1#~/.yarn/patches/ansi-color-npm-0.2.1-f7243d10a4.patch",
```
enisdenjo, line 72: Why are we patching the ansi-color package?

package.json
```diff
@@ -66,10 +66,10 @@
     "@graphql-mesh/types": "0.104.4",
     "@graphql-mesh/utils": "0.104.4",
     "@graphql-tools/delegate": "workspace:^",
-    "@opentelemetry/exporter-trace-otlp-http@npm:0.200.0": "patch:@opentelemetry/exporter-trace-otlp-http@npm%3A0.200.0#~/.yarn/patches/@opentelemetry-otlp-exporter-base-npm-0.200.0-6fb25211c7.patch",
-    "@opentelemetry/otlp-exporter-base@npm:0.200.0": "patch:@opentelemetry/otlp-exporter-base@npm%3A0.200.0#~/.yarn/patches/@opentelemetry-otlp-exporter-base-npm-0.200.0-6fb25211c7.patch",
-    "@opentelemetry/resources@npm:2.0.0": "patch:@opentelemetry/resources@npm%3A2.0.0#~/.yarn/patches/@opentelemetry-resources-npm-2.0.0-f20376a5f9.patch",
+    "@opentelemetry/otlp-exporter-base@npm:0.202.0": "patch:@opentelemetry/otlp-exporter-base@npm%3A0.202.0#~/.yarn/patches/@opentelemetry-otlp-exporter-base-npm-0.202.0-f6f29c2eeb.patch",
+    "@rollup/plugin-node-resolve@npm:^15.2.3": "patch:@rollup/plugin-node-resolve@npm%3A16.0.1#~/.yarn/patches/@rollup-plugin-node-resolve-npm-16.0.1-2936474bab.patch",
```
enisdenjo, line 70: Why are we patching rollup?

enisdenjo (changes requested):
- Please explain all patches in detail in DEPS_RESOLUTIONS_NOTES.md

internal/e2e/src/tenv.ts
EmrysMyrddin: Concat nothing, it's a way to make a copy of the array, so that when we push to it we doesn't modify the array provided in configuration. This is to allow passing the same array in each tests.

internal/testing/src/getLocalhost.ts
EmrysMyrddin: Yes, it shouldn't have been comited. You solution with pipelogs is working well. I will reverst changes

package.json
EmrysMyrddin: I needed to patch it so that it gives priority to the target conditions we add in configuration.
Otherwise, it picks the first condition matching, which doesn't work for OTEL packages.
This allows to avoid patching most of the OTEL packages.

package.json
EmrysMyrddin: Because it contains a legacy hex character syntax, which is not allowed in strict, which we are using in Hive Gateway.
This is used by the jaeger tracer (and by extension OTEL `NodeSDK`).

internal/e2e/src/tenv.ts
enisdenjo: Hmm, we should not be mutating this array, do we? Should we not?

enisdenjo (approved):
Aside from this one question from tenv.ts, this is looking good!

internal/e2e/src/tenv.ts
EmrysMyrddin: In the following code, we are adding default volumes, so it makes sense actually.
And my usage with a reusable env config doesn't seems to be that used since I'm the first one to have the issue ^^'

EmrysMyrddin: We can do it this way if you find it more conventional or obvious that we are doing a copy
```suggestion
            [...runner?.docker?.volumes ?? []];
```

enisdenjo: > In the following code, we are adding default volumes, so it makes sense actually.

But is it a problem? Do we get duplicate entries between test runs or?

EmrysMyrddin: Most of the time we don't.

Because most of the time, the config is not shared by multiple call to the gateway setup function: 
```ts
const instance = await gateway({
  //...
  runner: {
    docker: {
      volumes: [
        //... some additional volumes
      ]
    }
  }
})
```

But in my case, I've made a global config that I re-use:
```ts
const runner = { docker: { volumes: [...] } }

it('should ...', () => {
  const instance = await gateway({ runner })
  //...
})

it('should ...', () => {
  const instance = await gateway({ runner })
})
```

And that's the point of my modification. It no longer mutate directly the configuration array. It makes a new array each time with the apriopriate volumes (the one from the config + the default ones).

EmrysMyrddin: I've made change so that it's more obvious what I'm doing here. And it's more coherent with the rest of the code.
I create an empty volumes array and progressively fill it with volumes according to config.

enisdenjo: Perfect, great job! We love clarity. :D

EmrysMyrddin:
Rebasing and waiting for CI, then I merge !

PR #1231, opened 2025-06-12
feat(logger): add LogTape writer

XiNiHa:
**NOTE: shouldn't be merged until all TODOs are resolved**

This PR adds the writer implementation for [LogTape](https://logtape.org) in Hive Logger.

### DIFF
--- a/packages/logger/src/writers/logtape.ts
+++ b/packages/logger/src/writers/logtape.ts
@@ -1,0 +1,39 @@
+import { getLogger, Logger as LogTapeLogger } from "@logtape/logtape";
+import { LogLevel } from "../logger";
+import { Attributes } from "../utils";
+import { LogWriter } from "./common";
+
+export interface LogTapeLogWriterOptions {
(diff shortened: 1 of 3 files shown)


packages/logger/src/writers/logtape.ts
```diff
@@ -0,0 +1,36 @@
...
+	constructor(public options: LogTapeLogWriterOptions) {
+		this.#logTapeLogger = getLogger(this.options.category ?? ["graphql-hive"]);
+	}
+
+	write(
+		level: LogLevel,
+		attrs: Attributes | null | undefined,
+		msg: string | null | undefined,
+	): void {
+		const log = this.#logTapeLogger[level];
```
dahlia: The method should be bound:

```suggestion
		const log = this.#logTapeLogger[level].bind(this.#logTapeLogger);
```

enisdenjo (changes requested):
@XiNiHa there are suddenly a LOT of changes. Can you please rebase on v2 again so that we can merge this?

PR #1293, opened 2025-07-01
feat(opentelemetry): add log writer

EmrysMyrddin:
Add a Hive Logger compatible Log Writer for OpenTelemetry, with similar configuration API than `openTelemetrySetup`.

```ts
  logger: new Logger({
    writers: [new OpenTelemetryLogWriter({
      // Either:
      logger: logs.getLogger('my-custom-logger'), // Based on a pre-registered LoggerProvider
      // or
      processors: [...], // A list of custom providers
      // or
      exporter: new OTLPLogExporter({ url: '...' }), // An exporter that will be wrapped in a BatchedLogProcessor
      batching: ..., // optional batching config,
      // or
      console: true, // adds an stdout exporter without batching,
    })]
  }),
})
```


packages/plugins/opentelemetry/src/log-writer.ts
```diff
@@ -0,0 +1,149 @@
...
+      provider.forceFlush();
+    }
+  }
+
+  write(
+    level: LogLevel,
+    attrs: Attributes | null | undefined,
+    msg: string | null | undefined,
+  ): void | Promise<void> {
+    const attributes = Array.isArray(attrs) ? { attrs } : (attrs ?? undefined);
```
enisdenjo: I would still prefer the spread here avoidining unnecessary nesting.

```suggestion
    const attributes = Array.isArray(attrs) ? { ...attrs } : (attrs ?? undefined);
```

enisdenjo (changes requested):
Hive Logger is a dependency of the OTEL plugin, but is not listed in the package.json.

packages/plugins/opentelemetry/src/log-writer.ts
EmrysMyrddin: If you think it's better yes I can :-)

It's part of the open question I had in the PR body

enisdenjo: Yep, lets go ahead and spread it. 👍

enisdenjo (approved):
Looking good, great work!

PR #1300, opened 2025-07-05
feat(opentelemtry): Add CLI options for basic setup

EmrysMyrddin:
This PR aims to improve DX and getting started experience around Traces reporting. It is adding some CLI options for basic setup, and configuration helper for future Hive Tracing tool when using a config file

## CLI Options

1. Basic setup with a simple OTLP exporter over HTTP
```bash
hive-gateway supergraph --opentelemetry http://example.com/traces/v1
```

2. OTLP with grpc
```bash
hive-gateway supergraph --opentelemetry http://example.com/traces/v1 --opentelemtry-exporter-type otlp-grpc
```

3. Hive Tracing
```bash
hive-gateway supergraph --hive-target "..." --hive-trace-access-token "..."
```

4. With both Hive Tracing and traditional OTEL
```bash
hive-gateway supergraph --opentelemetry http://example.com/traces/v1 --hive-target "..." --hive-trace-access-token "..."
```

5. With both hive Tracing and Usage report
```bash
hive-gateway supergraph --hive-target --hive-access-token "..." --hive-trace-access-token "..."
```
The idea is that later, once hive tracing becomes stable, only `--hive-target` and `--hive-access-token` will be needed to enable both tracing and usage report at once.

## Helper in config file

You can now easily setup future Hive Tracing feature in configuration file or programatic usage

```ts
import { hiveTracingSetup } from '@graphql-mesh/plugin-opentelemetry/setup'

hiveTracingSetup({
  accessToken: '...',
  target: '...',
});
```


n1ru4l:
Seems like there is still a few parameters missing from the defined list:

```
Root Span Attributes:

hive.target_id          <string> # the id of the target (supergraph id or slug)

hive.graphql.error.count:        <number>      # amount of errors in operation
hive.graphql.error.codes:        <string[]>    # list of all data.errors.extensions.code strings
hive.graphql.operation.document: <string>      # the executable operation document
hive.graphql.operation.name:     <string>
hive.graphql.operation.type:     <query|mutation|subscription>
hive.graphql.operation.hash:     <string>      # hash as calulated to information below

hive.subgraph.names:         <string[]>    # contains subgraphs that were involved in execution

hive.client.name:            <string>
hive.client.version:         <string>

http.status_code:       <number>
http.method:            <string>
http.host:              <string>
http.route:             <string>
http.url:               <string>
```

```
"subgraph.execute ..." Span Attributes:

hive.target_id          <string> # the id of the target (supergraph id or slug)
hive.graphql.error.count:        <number>      # amount of errors in operation
hive.graphql.error.codes:        <string[]>    # list of all data.errors.extensions.code strings
hive.graphql.operation.document: <string>      # the executable operation document
hive.graphql.operation.name:     <string>
hive.graphql.operation.type:     <query|mutation|subscription>

hive.subgraph.name:         <string>    # contains subgraphs that were involved in execution
```

packages/plugins/opentelemetry/package.json
```diff
@@ -52,6 +52,7 @@
     "graphql": "^15.9.0 || ^16.9.0"
   },
   "dependencies": {
+    "@graphql-hive/core": "0.13.0-alpha-20250707104821-802de2e836e8fc0fb59533a3c4c63cf53e9fc33c",
```
enisdenjo: Do we want the alpha release here?

enisdenjo (changes requested):
You're doing great work! Just that one remark. I like the new tests.

packages/plugins/opentelemetry/package.json
EmrysMyrddin: No, we will have to wait for this PR to be merged: https://github.com/graphql-hive/console/pull/6876

packages/plugins/opentelemetry/src/spans.ts
```diff
@@ -127,10 +141,35 @@ export function setParamsAttributes(input: {
...
+  if (!typeInfos.has(input.schema)) {
+    typeInfos.set(input.schema, new TypeInfo(input.schema));
+  }
+  const typeInfo = typeInfos.get(input.schema);
+
+  return hashOperation({
+    documentNode: input.document,
+    operationName: input.operationName ?? null,
+    schema: input.schema,
+    variables: input.variableValues ?? null,
```
n1ru4l: Using the variables as part of the hash is opt-in:

https://github.com/graphql-hive/console/blob/92756d8d0c5e13333229290a81f7cc3189a9df80/packages/libraries/core/src/client/types.ts#L132-L137

EmrysMyrddin: Oh, I didn't noticed this. So you think there should be an option or I should just not pass it at all and let users override hashing function if they want to enable it ?

n1ru4l: For now I think we can just omit them, the feature is not stable yet anyways. Once we have people complain, we will make it configurable!

packages/plugins/opentelemetry/src/attributes.ts
```diff
@@ -21,8 +21,12 @@ export {
 export const SEMATTRS_GRAPHQL_DOCUMENT = 'graphql.document';
 export const SEMATTRS_GRAPHQL_OPERATION_TYPE = 'graphql.operation.type';
 export const SEMATTRS_GRAPHQL_OPERATION_NAME = 'graphql.operation.name';
+export const SEMATTRS_GRAPHQL_OPERATION_HASH = 'graphql.operation.hash';
```
n1ru4l: Do you think this is a good idea to use this for the hive hash? If there will be an official algorithm this would clash. So I would keep this attribute Hive specific `hive.graphql.operation.hash`

packages/plugins/opentelemetry/src/attributes.ts
```diff
@@ -21,8 +21,12 @@ export {
...
 export const SEMATTRS_GRAPHQL_ERROR_COUNT = 'graphql.error.count';
+export const SEMATTRS_GRAPHQL_ERROR_CODES = 'graphql.error.codes';
```
n1ru4l: These are also non-standard. Is it like the `graphql.` prefix is reserved for future spec adjustments? In that case I would also prefer using `hive.graphql` as the prefix? While a chance of a conflict seems low, I want to avoid it if possible.

packages/plugins/opentelemetry/src/attributes.ts
```diff
@@ -21,8 +21,12 @@ export {
...
 
 // Gateway-specific attributes
 export const SEMATTRS_GATEWAY_UPSTREAM_SUBGRAPH_NAME =
   'gateway.upstream.subgraph.name';
+export const SEMATTRS_GATEWAY_OPERATION_SUBGRAPH_NAMES =
+  'gateway.operation.subgraph.names';
```
n1ru4l: Where are these specified?

packages/plugins/opentelemetry/src/attributes.ts
EmrysMyrddin: At otel-wg level, the idea for now is to have some sort of "hash" or "id". The content, how it is computed, is let to the implementor.

But yes, we can :-) I'm just afraid of the compute cost of this hash when not needed.

I mean, we want a hash for sure. So I have to find a way to reliably tell if the hash is a hive hash or not

packages/plugins/opentelemetry/src/spans.ts
```diff
@@ -311,18 +372,20 @@ export function createSubgraphExecuteSpan(input: {
...
+    input.executionRequest.document,
+    input.executionRequest.operationName,
+  );
+
   const span = input.tracer.startSpan(
     `subgraph.execute (${input.subgraphName})`,
     {
       attributes: {
-        [SEMATTRS_GRAPHQL_OPERATION_NAME]: input.executionRequest.operationName,
+        [SEMATTRS_GRAPHQL_OPERATION_NAME]: operation.name?.value ?? 'Anonymous',
```
n1ru4l: Since the following is possible, it seems like an empty string indicating the absence would be a better fit.

```
query Anonymous {
  __typename
}
```

packages/plugins/opentelemetry/src/attributes.ts
EmrysMyrddin: They are not for now, I have to add them to documentation

packages/plugins/opentelemetry/src/attributes.ts
EmrysMyrddin: We can prefix with hive and remove the prefix once it is standardize ? It just means a "breaking change"

packages/plugins/opentelemetry/src/spans.ts
EmrysMyrddin: Yes you are right. I kept it for backward compatibility at first, but now that we decided to cut a v2, I will change that. You prefer an empty string over the absence of the attribute ? It's the default behavior for optional attributes, the are just not reported if the value is undefined.

packages/plugins/opentelemetry/package.json
n1ru4l: `@graphql-hive/core@0.13.0` got released: https://github.com/graphql-hive/console/pull/6869

packages/plugins/opentelemetry/src/spans.ts
n1ru4l:
```suggestion
    variables: null,
```

packages/plugins/opentelemetry/src/attributes.ts
n1ru4l: We only need to add it for now when using the "hive" mode. It seems like the "generic" hash is something that is still in discussion and not anywhere near from being finalized according to the working group notes.

packages/plugins/opentelemetry/src/attributes.ts
n1ru4l: I see https://opentelemetry.io/docs/specs/semconv/registry/attributes/gateway/ does not exist.

Not sure how often attributes and namespaces are made "official", but I personally would always pick a prefix that makes clashes in the future very unlikely.

I am fine with using these though if you think it is the best choice.

packages/plugins/opentelemetry/src/attributes.ts
n1ru4l: From my platform perspective, we need to be aware if existing attributes that we rely on are changing as it will require adjusting our processing configuration as we aggregate and index traces.

While it seems unlikely that `graphql.error.count` and `graphql.error.codes` would differ drastically, from how we implement them today, if such a change would occur it would be annoying from the hive console pre-processing perspective, but not impossible.

packages/plugins/opentelemetry/src/hive-span-processor.ts
```diff
@@ -0,0 +1,197 @@
...
+  private traceStateById: Map<string, TraceState> = new Map();
+  private processor: SpanProcessor;
+
+  constructor(config: HiveTracingSpanProcessorOptions) {
+    if (config.processor) {
+      this.processor = config.processor;
+    } else {
+      this.processor = new BatchSpanProcessor(
+        new OTLPTraceExporter({
+          url: `https://app.graphql-hive.com/opentelemetry/v1/traces`,
```
n1ru4l: This needs to be configurable via a CLI parameter as I need to set it to localhost in local development.
I would propose something like `--hive-trace-endpoint`?

The default production value will be `https://api.graphql-hive.com/otel/v1/traces`.

n1ru4l: https://github.com/graphql-hive/gateway/pull/1334

packages/plugins/opentelemetry/src/attributes.ts
EmrysMyrddin: I'm prefixing for now. If it get standardized, we will either update Console to be based on the standard name, or the custom processor to simply rename the attribute (and make the adaptations if needed)

packages/plugins/opentelemetry/package.json
```diff
@@ -52,6 +52,7 @@
     "graphql": "^15.9.0 || ^16.9.0"
   },
   "dependencies": {
+    "@graphql-hive/core": "0.13.0",
```
enisdenjo: Why pin the version?
```suggestion
    "@graphql-hive/core": "^0.13.0",
```

enisdenjo: https://github.com/graphql-hive/gateway/pull/1300/commits/f4028536738659cad2bfaf87b3dcc44da357732b

PR #1310, opened 2025-07-09
feat(gateway): Export `withState` util from GraphQL Yoga

EmrysMyrddin:


packages/runtime/src/index.ts
```diff
@@ -17,3 +17,4 @@ export { useUpstreamRetry } from './plugins/useUpstreamRetry';
 export { useUpstreamTimeout } from './plugins/useUpstreamTimeout';
 export { getGraphQLWSOptions } from './getGraphQLWSOptions';
 export * from './getDefaultLogger';
+export { withState } from 'graphql-yoga';
```
enisdenjo: Since envelop is the originating package and the runtime depends on it, we should re-export from there and not create a chain of dependencies.

```suggestion
export { withState } from '@envelop/core';
```

EmrysMyrddin: Sure yes, good idea, it will be easier to update.

EmrysMyrddin:
@enisdenjo Do you have an idea why the test started failing ? Did we modified a default or something in Yoga ?

enisdenjo (approved):
Looking good!

EmrysMyrddin:
Seems it was flakey ^^'

ardatan:
@EmrysMyrddin @enisdenjo This change can fix it, Node's creating async iterator logic seems heavy
https://github.com/ardatan/whatwg-node/pull/2578

enisdenjo (approved):

Issue #1311, opened 2025-07-10
Race condition in header propagation for multi-field federation queries causes subgraphName to be undefined


rowanparker:
There is a race condition in the header propagation system when executing multi-field GraphQL queries that span multiple subgraphs
  in federation mode. The `fromClientToSubgraphs` function receives `subgraphName: undefined` during parallel execution, causing
  authentication failures for queries with multiple top-level fields.

 ####  Steps to Reproduce

  1. Set up a federated GraphQL gateway with multiple subgraphs
  2. Configure header propagation using `fromClientToSubgraphs`
  3. Execute a query with multiple top-level fields from different subgraphs:
```graphql
  query MultiFieldQuery {
    userProfile {    # From UserService subgraph
      id
      name
    }
    products {       # From ProductService subgraph
      id
      title
    }
  }
```
  4. Observe that `subgraphName` parameter is `undefined` in the `fromClientToSubgraphs` function

 #### Expected Behavior

  The `fromClientToSubgraphs` function should receive the correct `subgraphName` for each subgraph request, allowing for
  subgraph-specific header propagation.

####  Actual Behavior

  - Single-field queries work correctly (subgraphName is properly set)
  - Multi-field queries fail with `subgraphName: undefined`
  - Authentication headers are not properly propagated to subgraphs

####  Environment

  - @graphql-hive/gateway: 1.15.4
  - @graphql-mesh/fusion-runtime: 0.11.17
  - Node.js: 18+

 #### Root Cause Analysis

  Based on source code analysis, this is a race condition between two hook execution flows:

  1. Header Propagation Flow

  File: `@graphql-hive/gateway-runtime/dist/index.js:1183`
  const subgraphName = executionRequest && subgraphNameByExecutionRequest.get(executionRequest);

  2. Subgraph Execution Flow

  File: `@graphql-mesh/fusion-runtime/dist/index.js:929`
  subgraphNameByExecutionRequest.set(executionRequest, subgraphName);

  The Race Condition

  1. The `subgraphNameByExecutionRequest` WeakMap is populated in `onSubgraphExecute()`
  2. Header propagation tries to read from this WeakMap in `onFetch()`
  3. During federation parallel execution, onFetch() is called BEFORE `onSubgraphExecute()`
  4. This causes `subgraphNameByExecutionRequest.get()` to return `undefined`

  Why Single-field Queries Work

  Single-field queries use sequential execution where timing is deterministic and `onSubgraphExecute()` populates the WeakMap before
[...]

PR #1313, opened 2025-07-10
fix(gateway-runtime): Add `subgraphName` to `ExecutionRequest`

EmrysMyrddin:
Fixes a bug preventing `subgraphName` to be passed correctly to `fromClientToSubgraphs` when subgraph requests gets batched into a single execution request.

The root cause is more global: the gateway is using a weak map to keep track of the request execution's subgraph name. Which is why it breaks when the execution request is overriden. It was probably also impacting retry logic.

The fix is to move subgraph name directly to the `ExecutionRequest` type. This way, it is easier to track and discover for newcomers.

Related to #1311.

### DIFF
--- a/.changeset/big-dolls-invent.md
+++ b/.changeset/big-dolls-invent.md
@@ -1,0 +1,10 @@
+---
+'@graphql-mesh/fusion-runtime': minor
+---
+
+Breaking Change: Removed `subgraphNameByExecutionRequest` weak map. Subgraph name is now stored in the execution request itself.
+
+```diff
+- const subgraphName = subgraphNameByExecutionRequest.get(executionRequest)
++ const subgraphName = executionRequest.subgraphName
+```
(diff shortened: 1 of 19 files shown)


EmrysMyrddin:
Yes, it is intentional. It was on main at the beginning, but the fix contains a breaking change (the removal of the `subgraphNameByExecutionRequest` map)

ardatan:
It is a breaking change in an internal package actually.

EmrysMyrddin:
@ardatan Yes, it's kind of supposed to be internal, but it is exported, and the only way to get the subgraph name in the `onFetch` hook, so still possible a motivated user found it out :-P

ardatan:
> @ardatan Not from the main package, but one can just import @graphql-mesh/fusion-runtime to have access to it.

If that's the case, we can just bump `@graphql-mesh/fusion-runtime` to a next minor with this breaking change because it is still v0.

EmrysMyrddin:
That's true, but anyways, I think it's ok to have it in v2 ? We are close to release, so I think it's ok to wait ?
Or it's because you want to fix it in v1 for people that will not update to v2 immediatly ?

ardatan (approved):

PR #1335, opened 2025-07-16
Rename attributes

n1ru4l:


EmrysMyrddin:
I've already updated my PR with those changes.

Warning: In my implementation, the operation name attribute is just omitted for anonymous operations. In this case the operation span is named `graphql.operation` without any postfix.

PR #1343, opened 2025-07-18
fix(opentelemetry): Patch otel sdk for span start time with nanosecond precision 

EmrysMyrddin:
Related: https://github.com/open-telemetry/opentelemetry-js/pull/5804


packages/plugins/opentelemetry/package.json
```diff
@@ -71,7 +71,7 @@
     "@opentelemetry/resources": "^2.0.1",
     "@opentelemetry/sdk-logs": "^0.203.0",
     "@opentelemetry/sdk-node": "^0.203.0",
-    "@opentelemetry/sdk-trace-base": "^2.0.1",
+    "@opentelemetry/sdk-trace-base": "patch:@opentelemetry/sdk-trace-base@patch%3A@opentelemetry/sdk-trace-base@npm%253A2.0.1%23~/.yarn/patches/@opentelemetry-sdk-trace-base-npm-2.0.1-ebe4f8e34e.patch%3A%3Aversion=2.0.1&hash=212481#~/.yarn/patches/@opentelemetry-sdk-trace-base-patch-0b7dbf6a30.patch",
```
enisdenjo, line 74: Publishing this will break installs. Patches are only for local things. If patching the library is the only way to go, then we need to handle it differently - on Hive Console side.

P.S. Node has only millisecond precision, so we cannot expect it to give out accurate nanoseconds.

EmrysMyrddin: Oh yes, I should only add this to the root package.json.

It's more an experiment at this point for @n1ru4l 

We can ship this in the binary, but not sure we should have such behavior diff between binary and node usage.

I'm trying to make a PR on opentelemetry side. Perhaps I can publish a fork ?

enisdenjo: I think pushing this PR to OTEL is a good way to go. But we dont have to do all that just for one change, do you think they'd release soon or would it take months?

EmrysMyrddin: Based on other PR dates, it takes loooong time to merge things, even for member of the repo...

enisdenjo: And we cannot handle it in Hive Console @n1ru4l? I can imagine others using Hive Console's observability features having the same issues, especially if based on OTEL's libs.

EmrysMyrddin: It can't be on Hive Console side, since it's the data producer that has the "bug".
It's span creation that uses a low precision timestamp...

A workaround for us waiting for a proper upstream fix would be to manually provide the start and end times. This way, we are sure that at least our spans have correct precision.

enisdenjo:
This one needs proper thinking and potentially fixes upstream. Let's rethink it and revisit when ready.

PR #1385, opened 2025-08-11
fix(@graphql-mesh/transport-http): infinite loop on subscriptions without kind

EmrysMyrddin:
HTTP transport loading is causing infinite loop when `subscriptions` option is provided with the `http` kind (or if `kind` is not specified since it defaults to `http`).

This is caused by the fact we pass all parent transport entry options to the subscription specific transport. In the case of the `http` transport, it was causing an infinite loop since a new subscriptions specific transport was instanciated every time because `subscriptions` option was always present.

This PR fixes this by explicitly removing the `subscriptions` attribute from the options passed to the subscription transport builder.


EmrysMyrddin:
Yeah I will try to add a unit test

PR #1395, opened 2025-08-19
Async compatible Hive PubSub, Redis and NATS PubSub, EDFS

enisdenjo:
#snapshot-npm-tag "next"

Ref GW-448

In order to support advanced PubSub mechanisms, like implementing Redis, the PubSub interface needs to change and start supporting promises in subscribe and publish, a necessary change to add guarantees and proper error handling. While at it, we refactored and dropped legacy methods and functions, moderanising and refreshing the API adding support for advanced primitives.

[See changelog for more information and migration.](https://github.com/graphql-hive/gateway/blob/dist-sub-web/.changeset/hip-suits-dress.md)

### TODO
- [x] Changeset
- [x] Pubsub migration
- [x] Gateway migration
- [x] Documentation website
- [x] Tests, also for Redis and NATS PubSub
- [x] Document EDFS
- [x] E2E example (we cannot convert it to a regular example because the examples converter does not support containers)
- [x] ~~The 'destroy' event will be published to all subscribers too (from legacy Mesh)~~
  Won't be an issue because it is [used only in legacy Mesh v0](https://github.com/search?q=repo%3Aardatan%2Fgraphql-mesh+%22.publish%28%27destroy%22++NOT+path%3A*.spec.*&type=code)
- [x] https://github.com/ardatan/graphql-mesh/pull/8716
  - [x] Use stable versions of Mesh packages (#1408)
- [x] https://github.com/ardatan/graphql-mesh/pull/8736
  - [x] Use stable versions and revert https://github.com/graphql-hive/gateway/pull/1395/commits/d81431bfa96bf4ea1f1c4593f4a1e684aeb712c5


packages/runtime/src/plugins/useWebhooks.ts
```diff
@@ -24,15 +24,13 @@ export function useWebhooks({
     See documentation: https://the-guild.dev/docs/mesh/pubsub`);
   }
   return {
-    onRequest({ request, url, endResponse, fetchAPI }) {
-      const eventNames = pubsub.getEventNames();
-      if ((eventNames as string[]).length === 0) {
-        return;
-      }
+    async onRequest({ request, url, endResponse, fetchAPI }) {
+      const topics = await pubsub.subscribedTopics();
```
ardatan: Is it possible to avoid `async/await` here? This is not always promise right?

packages/pubsub/src/mem.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (listeners!.size === 0) {
+        this.#subscribers.delete(topic);
+      }
+    };
+    listeners.set(listener, unsubscribe);
+
+    return unsubscribe;
+  }
+
+  public async dispose() {
```
ardatan: This is sync here?

enisdenjo: https://github.com/graphql-hive/gateway/pull/1395/commits/756cd806e1c6f15fecfcdc08889386011844c11f

packages/runtime/src/plugins/useWebhooks.ts
enisdenjo: https://github.com/graphql-hive/gateway/pull/1395/commits/1374f636c439d91725ef6761f936f8160986a48b

Issue #1404, opened 2025-08-26
@graphql-tools/stitch doesn't preserve @oneOf


jwhitaker-gridcog:
Hey guys, thanks for maintaining the tools herein.

We've noticed that `stitchSchemas` doesn't seem to preserve `@oneOf`. Repro:

```ts
import {stitchSchemas} from '@graphql-tools/stitch';
import {buildSchema, printSchema} from 'graphql';

const schemaSrc = `
type Query {
  findPerson(in: FindPersonInput): Person
  findPersonById(id: String): Person @deprecated(reason: "use findPerson")
}

type Person {
  Name: String
  ID: String
}

input FindPersonInput @oneOf {
  byId: String
  byName: String
}
`;

const schema = buildSchema(schemaSrc);
const stitchedSchema = stitchSchemas({subschemas: [schema]});

console.log(printSchema(schema));
console.log('-----------')
console.log(printSchema(stitchedSchema));
```

Output:
```graphql
type Query {
  findPerson(in: FindPersonInput): Person
  findPersonById(id: String): Person @deprecated(reason: "use findPerson")
}

type Person {
  Name: String
  ID: String
}

input FindPersonInput @oneOf {
  byId: String
  byName: String
}
-----------
type Query {
  findPerson(in: FindPersonInput): Person
  findPersonById(id: String): Person @deprecated(reason: "use findPerson")
}

type Person {
  Name: String
  ID: String
}

input FindPersonInput { # <--- missing @oneOf!
  byId: String
  byName: String
}
```

Any reason why this is the case? I've included a `@deprecated` as a demo of another built-in directive that seems to work out of the box.

Cheers!

package.json:
```json
{
  "name": "stitchbug",
  "version": "1.0.0",
  "description": "",
  "license": "ISC",
  "dependencies": {
    "@graphql-tools/stitch": "^9.4.28",
    "graphql": "^16.11.0"
  },
  "devDependencies": {
    "tsx": "^4.20.5",
    "typescript": "^5.9.2"
  }
}
```

PR #1405, opened 2025-08-26
fix(stitch): handle `@oneOf`s in subschemas

ardatan:
Fixes https://github.com/graphql-hive/gateway/issues/1404

### DIFF
--- a/packages/stitch/src/mergeCandidates.ts
+++ b/packages/stitch/src/mergeCandidates.ts
@@ -64,6 +64,9 @@ export function mergeCandidates<TContext = Record<string, any>>(
   typeMergingOptions?: TypeMergingOptions<TContext>,
 ): GraphQLNamedType {
   const initialCandidateType = candidates[0]?.type;
+  if (candidates.length === 1) {
+    return initialCandidateType!;
+  }
   if (
     candidates.some(
       (candidate) =>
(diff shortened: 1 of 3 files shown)


jwhitaker-gridcog:
hey @enisdenjo , any chance you could have a look at this? sorry to nag :)

enisdenjo (approved):

packages/stitch/src/mergeCandidates.ts
```diff
@@ -64,6 +64,9 @@ export function mergeCandidates<TContext = Record<string, any>>(
   typeMergingOptions?: TypeMergingOptions<TContext>,
 ): GraphQLNamedType {
   const initialCandidateType = candidates[0]?.type;
+  if (candidates.length === 1) {
+    return initialCandidateType!;
```
namoscato, line 68: FYI presumably an edge case, but this optimization seemed to result in a breaking change. Previously when stitching together a single `subschemas`, root fields would always be normalized (i.e. `query_root` would get renamed to `Query`). Now that is no longer the case.

ardatan: Could you create a seperate issue with a reproduction? Thanks!

namoscato: https://github.com/graphql-hive/gateway/issues/1537

PR #1419, opened 2025-08-29
Deduplicate inflight requests

ardatan:


enisdenjo (commented):
A few things
- Lets base this PR on v2
- Make it configurable in the config, we can have it enabled by default
- Add some documentation, blog post is already ready

PR #1421, opened 2025-09-01
feat(opentelemetry): add global API for context utils

EmrysMyrddin:


packages/plugins/opentelemetry/tests/utils.ts
```diff
@@ -75,26 +73,24 @@ export async function buildTestGateway(
...
-          otelPlugin,
-          ...(options.plugins?.(ctx, otelPlugin) ?? []),
+          useOpenTelemetry({
+            ...ctx,
+            ...options.options,
+          }),
+          ...(options.plugins?.(ctx) ?? []),
         ];
       },
-      logging: false,
```
enisdenjo, line 91: Do we need logging here?

packages/plugins/opentelemetry/src/api.ts
```diff
@@ -0,0 +1,63 @@
+import { Logger } from '@graphql-hive/logger';
+import { context, Tracer } from '@opentelemetry/api';
+import type { OpenTelemetryPluginUtils } from './plugin';
+
+export * from '@opentelemetry/api';
+
+export type HiveAPI = Omit<OpenTelemetryPluginUtils, 'tracer'> & {
+  tracer?: Tracer;
```
enisdenjo: Will be set but potentially undefined.
```suggestion
  tracer: Tracer | undefined;
```

packages/plugins/opentelemetry/src/api.ts
```diff
@@ -0,0 +1,63 @@
...
+   *
+   * @param utils: The plugin instance with all the utility methods
+   * @param log: A Logger that will be used to warn in case of double registration. Leave it undefined or null to disable the log.
+   * @returns true if successful, false if the API was already registered
+   */
+  setPluginUtils(utils: OpenTelemetryPluginUtils, log?: Logger): boolean;
+  /**
+   * Unregister the current Hive OpenTelemetry plugin utility API if any is registered. No-op if no API was registered
+   */
+  disable(): void;
```
enisdenjo, line 29: "disable" implies that "enable" is possible, which it is not. I'd go with a more destructive name like `unregister`.

```suggestion
  unregister(): void;
```

packages/plugins/opentelemetry/src/api.ts
enisdenjo: You did great with commenting the rest of options, lets leave comment for this one too. What is it? When is it undefined?

packages/plugins/opentelemetry/tests/utils.ts
EmrysMyrddin: Oh no, test leftovers

packages/plugins/opentelemetry/src/api.ts
EmrysMyrddin: I agree, but it was to follow the convention of other OTEL api. Like `context.disable()` or `trace.disable()`

enisdenjo: Oh ok! Got it.

enisdenjo: As long as OTEL does not have .enable(), lets keep it.

PR #1446, opened 2025-09-05
Fix Windows CI

ardatan:


enisdenjo:
I have disabled the Win E2E CI in #1448.

PR #1480, opened 2025-09-10
docs: add root CHANGELOG linking to gateway

EmrysMyrddin:


CHANGELOG.md
```diff
@@ -0,0 +1 @@
+packages/gateway/CHANGELOG.md
```
ardatan, line 1: This doesn't seem working well;
https://github.com/graphql-hive/gateway/blob/docs/add-root-changelog/CHANGELOG.md

ardatan (changes requested):

CHANGELOG.md
EmrysMyrddin: Not sure why it doesn't work on this view, but it should be working.
We've done the same thing for `.nvmrc` which is a symbolic link to `.node-version`.

Not sure why it doesn't work anymore, seems like a Github bug... It does the same on the yoga repo where we are doing the same thing and where it used to work.

ardatan: Maybe it works probably it is a view bug in GH.

PR #1493, opened 2025-09-15
Inject config context into the server context before all plugins avoiding potential short-circuits

enisdenjo:
### TODO

- [x] Write tests making sure the logger is available even when not dealing with graphql requests

### DIFF
--- a/packages/runtime/src/plugins/useConfigInServerContext.ts
+++ b/packages/runtime/src/plugins/useConfigInServerContext.ts
@@ -1,4 +1,5 @@
 import { getHeadersObj } from '@graphql-mesh/utils';
+import { ServerAdapterPlugin } from '@whatwg-node/server';
 import { GatewayConfigContext, GatewayPlugin } from '../types';
 
 export interface ConfigInServerContextOptions {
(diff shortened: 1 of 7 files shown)


packages/runtime/src/plugins/useConfigInServerContext.ts
```diff
@@ -8,7 +9,7 @@ export interface ConfigInServerContextOptions {
 export function useConfigInServerContext({
   configContext,
 }: ConfigInServerContextOptions): GatewayPlugin {
-  return {
+  const configInServerContextPlugin: ServerAdapterPlugin = {
```
ardatan, line 12: Let's also deduplicate the logic here;
https://github.com/graphql-hive/gateway/blob/first-log/packages/runtime/src/createGatewayRuntime.ts#L1092
It causes to the (relatively expensive) object spread etc multiple times.

enisdenjo: it's not duplicate logic, the `context` in yoga is executed on graphql context building. for example, the websocket headers wont be available in onRequest (because it will never be invoked)

ardatan: Hmm still `configContext` is spreaded twice in this case(when onRequest is called), also `headers` is added twice, no?

enisdenjo: Did [this](https://github.com/graphql-hive/gateway/pull/1493/commits/dbd291df5f3e5a8600af276f532cb801e1bb8197), wdyt?

ardatan: That's also good but I meant this;
https://github.com/graphql-hive/gateway/commit/1776da4ac1de4c82924349c37616c6185b074ddc

ardatan: Have you checked the commit I shared?

PR #1533, opened 2025-09-19
fix(federation): keep the original subgraph reference while handling shared root resolution

ardatan:
Handle heavily nested shared root selection as in Grafbase's many-cases benchmark scenario correctly
It was an issue with the merging when the original subschema reference is lost


packages/federation/src/supergraph.ts
```diff
@@ -1676,11 +1680,26 @@ function mergeResults(results: unknown[], getFieldNames: () => Set<string>) {
...
+        for (const data of datas) {
+          // @ts-expect-error - we know it is there
+          const symbolValue = data?.[symbol];
+          if (symbolValue != null) {
+            mergedData[symbol] = symbolValue;
+            break;
+          }
+        }
+      }
+    }
```
EmrysMyrddin, line 1701: To avoid this, in the `withState` util, I'm copying property descriptors instead of attributes.

This way, you can copy everything, including properties that have a getter instead of an actual value

EmrysMyrddin (approved):

PR #1557, opened 2025-09-24
fix(delegate): apply type-merging in extension fields

ardatan:
Closes https://github.com/ardatan/graphql-tools/pull/4994

When there is a computed field in the extension field's return type, there was a bug causing the type merging skipped.
This PR solves that issue.

### DIFF
--- a/packages/delegate/src/resolveExternalValue.ts
+++ b/packages/delegate/src/resolveExternalValue.ts
@@ -157,7 +157,7 @@ function resolveExternalObject<TContext extends Record<string, any>>(
         ] of potentialMergedTypeInfo.targetSubschemas) {
           if (
             targetSubschemas.length &&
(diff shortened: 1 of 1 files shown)


packages/delegate/src/resolveExternalValue.ts
```diff
@@ -146,6 +146,33 @@ function resolveExternalObject<TContext extends Record<string, any>>(
...
+            subschema = sourceSubschema as SubschemaConfig<
+              any,
+              any,
+              any,
+              TContext
+            >;
+            mergedTypeInfo = potentialMergedTypeInfo;
+          }
+        }
+        break;
```
EmrysMyrddin, line 171: Why do we always break here? We don't check if `mergedTypeInfo` is actually not null to check remaining `possibleTypeNames`?

Issue #1082, opened 2025-04-28
Federation 2.7 - Progressive override


jdolle:
Progressive overrides are a safer way to handle migrations. This feature should be implemented in the gateway.

https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#progressive-override

PR #1624, opened 2025-10-24
Progressive Override

ardatan:
Ref GW-323
Closes https://github.com/graphql-hive/gateway/issues/1082

Support for Progressive Override
- Handles `percent(x)` usage by default
- Documentation -> https://github.com/graphql-hive/console/pull/7168
- Maybe tests for `percent(` usage

If you have labels like below;
```graphql
type Foo {
   # If label is active, it overrides `Foo.bar` from A, and use this subgraph not A.
   bar: String @override(from: "A", label: "my_custom_label")
}
```

```ts
defineConfig({
  // You can control the label through headers like below;
  progressiveOverride(label, context) {
     if (label === 'my_custom_label' && context.request.headers.has('x-use-my-custom-label')) {
        return true;
     }
     return false;
  }
})
```

### DIFF
--- a/.changeset/cool-wolves-sort.md
+++ b/.changeset/cool-wolves-sort.md
@@ -6,35 +6,8 @@
 '@graphql-tools/stitch': minor
 ---
 
-Progressive Override Implementation
+Progressive Override for Safer Field Migrations
 
-By default the label of `percent(x)` will work as a simple percentage chance to override the field from another subgraph.
-But sometimes you may want to have more control over when to override based on request context, headers, or environment variables in a feature flag style.
-
-Override labels like below;
-```graphql
-type Foo {
-   # If label is active, it overrides `Foo.bar` from A, and use this subgraph not A.
-   bar: String @override(from: "A", label: "my_custom_label")
-}
-```
-And now handled within the configuration property `progressiveOverride` like below:
-```ts
-defineConfig({
-  // You can control the label through headers like below;
-  progressiveOverride(label, context) {
-     if (label === 'my_custom_label') {
-        // Use the headers
-        return context.request.headers.get('use-my-custom-label') === 'true';
-        // Use environment variables
-        return process.env.USE_MY_CUSTOM_LABEL === 'true';
-        // Or any custom logic that returns a boolean
-        // This example makes it override 50% of the time
-        return Math.random() < 0.5;
-     }
-     return false;
-  }
-})
-```
+Introduces Progressive Override, allowing you to safely migrate fields between subgraphs using the @override directive with a label. Control the rollout using custom logic in the gateway (e.g., percentage, headers) or the built-in percent(x) label for gradual, incremental traffic migration.
 
 Detailed documentation can be found [here](https://the-guild.dev/graphql/hive/docs/gateway/other-features/progressive-override).


packages/federation/src/supergraph.ts
```diff
@@ -1219,6 +1275,64 @@ export function getStitchingOptionsFromSupergraphSdl(
...
+            fieldsConfig = mergedConfig.fields = {};
+          }
+          let fieldConfig = fieldsConfig[fieldInfo.field];
+          if (!fieldConfig) {
+            fieldConfig = fieldsConfig[fieldInfo.field] = {};
+          }
+
+          const label = fieldInfo.label;
+          // Extract 10 from percent(10) for example
+          const percentRegexp = /^percent\((\d{1,3})\)$/;
```
kamilkisiela: A valid value for percentage is, I think, up to 8 digits or something

ardatan: Hmm I guess they don't have a limit;
https://github.com/apollographql/router/blob/dev/apollo-router/src/plugins/progressive_override/mod.rs#L113
But they parse it as `float` so I should remove the limit too.

.changeset/cool-wolves-sort.md
```diff
@@ -0,0 +1,9 @@
+---
+'@graphql-mesh/fusion-runtime': minor
+'@graphql-tools/federation': minor
+'@graphql-tools/delegate': minor
+'@graphql-hive/gateway-runtime': minor
+'@graphql-tools/stitch': minor
+---
+
+Progressive Override Implementation
```
n1ru4l: This changeset could show an actual example and link to documentation

packages/federation/src/supergraph.ts
kamilkisiela: The "limit" is limited by the producer of supergraph (composition - longer numbers fail composition)

.changeset/cool-wolves-sort.md
ardatan: Added

packages/federation/src/supergraph.ts
```diff
@@ -1219,6 +1275,49 @@ export function getStitchingOptionsFromSupergraphSdl(
...
+            fieldsConfig = mergedConfig.fields = {};
+          }
+          let fieldConfig = fieldsConfig[fieldInfo.field];
+          if (!fieldConfig) {
+            fieldConfig = fieldsConfig[fieldInfo.field] = {};
+          }
+
+          const label = fieldInfo.label;
+          // Extract 10 from percent(10) for example
+          const percentRegexp = /^percent\((\d+)\)$/;
```
kamilkisiela: \d+ won’t pass floats

packages/federation/src/supergraph.ts
ardatan: So we don't need to limit it as it is right now, no?

packages/federation/src/supergraph.ts
ardatan: Fixed

.changeset/cool-wolves-sort.md
```diff
@@ -0,0 +1,40 @@
...
+
+By default the label of `percent(x)` will work as a simple percentage chance to override the field from another subgraph.
+But sometimes you may want to have more control over when to override based on request context, headers, or environment variables in a feature flag style.
+
+Override labels like below;
+```graphql
+type Foo {
+   # If label is active, it overrides `Foo.bar` from A, and use this subgraph not A.
+   bar: String @override(from: "A", label: "my_custom_label")
```
n1ru4l: What happens if I don't handle `my_custom_label ` in `progressiveOverride`, will it print a warning or will it just return `false` by default?

Should we do a warning to the console or not?

packages/federation/src/supergraph.ts
kamilkisiela: I do limit it to avoid weird floats

.changeset/cool-wolves-sort.md
n1ru4l: From https://github.com/graphql-hive/gateway/pull/1624/files#diff-39e339bb347f4e24b0d77dce69cac1c65fe8fdad7f90072ecb19c20b547c230dR25-R35 it seems like we encourage people to provide a default fallback value.

Maybe the default fallback value should be `undefined` and then we print a "Could not resolve label" warning in the console in this scenario? So there is some visibility for an unresolvable label?

PR #1625, opened 2025-10-24
Hive Gateway tester and testing utilities

enisdenjo:
```ts
import { createGatewayTester } from '@graphql-hive/gateway-testing';

{
  // will be disposed of on scope exit
  await using gw = createGatewayTester({
    subgraphs: [
      {
        name: 'helloer',
        schema: {
          typeDefs: /* GraphQL */ `
            type Query {
              hello: String
            }
          `,
          resolvers: {
            Query: {
              hello: () => 'world',
            },
          },
        },
      },
    ],
    // your gateway config for testing
    plugins: () => [
      // your plugins for testing
    ],
  });

  // gateway is running in supergraph mode

  const result = await gw.execute({
    query: /* GraphQL */ `
      query {
        hello
      }
    `,
  });

  // the complete lifecycle of the gateway has been run

  console.log(result); // { data: { hello: 'world' } }
}
```

### TODO

- [x] Supergraph mode
- [ ] Subgraph mode
- [x] Proxy mode
- [x] Explicit resource management (`await using gw = createGatewayTester()`)
- [x] `dispose()` as alternative to `await using`
- [x] https://github.com/graphql-hive/console/pull/7180
- [x] Refactor internal tests to use the tester
  - A good chunk was refactored, but not all to avoid wasting time

### DIFF
--- a/packages/plugins/prometheus/tests/prometheus.spec.ts
+++ b/packages/plugins/prometheus/tests/prometheus.spec.ts
@@ -3,7 +3,6 @@ import { createGatewayTester } from '@graphql-hive/gateway-testing';
 import InMemoryLRUCache from '@graphql-mesh/cache-inmemory-lru';
 import { getUnifiedGraphGracefully } from '@graphql-mesh/fusion-composition';
 import { createDefaultExecutor } from '@graphql-mesh/transport-common';
-import { isDebug } from '@internal/testing';
 import { createSchema } from 'graphql-yoga';
 import { Registry, register as registry } from 'prom-client';
 import { beforeEach, describe, expect, it } from 'vitest';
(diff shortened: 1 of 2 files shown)


packages/testing/src/createGatewayTester.ts
```diff
@@ -0,0 +1,143 @@
...
+            info,
+          );
+        }),
+        ...(config.plugins?.(ctx) || []),
+      ],
+    });
+  }
+
+  const runtimeExecute = buildHTTPExecutor({
+    fetch: runtime.fetch,
```
ardatan: Let's add `endpoint` here to make Bun happy.
https://github.com/graphql-hive/gateway/actions/runs/18780606631/job/53585575434?pr=1625#step:4:6249
```suggestion
    endpoint: 'http://localhost/graphql',
    fetch: runtime.fetch,
```

PR #1636, opened 2025-10-27
feat(opentelemetry): add schema coordinate to error events

EmrysMyrddin:
# Context

Today, errors as reported as classic OTEL Exceptions. Following the work of the GraphQL-OTEL working group (https://github.com/graphql/otel-wg/pull/50) on GraphQL error reporting, Exceptions are not the right model fit.

Exception are for uncaught runtime errors. GraphQL Errors in the other hand are user facing business errors, with much more metadata than the traditional `message`, `type` and `stacktrace`.

Related to GW-501

# Solution

This PR migrates GraphQL errors to OTEL Event. This is following upcomging OTEL standard for GraphQL, and allows to add more metadata to it.

Errors are also attached to the operation root span, for easier filtering in dashboards.

Normal JS Error are still recorded as OTEL Exceptions, only actual `GraphQLError` found in the result, or returned by `validation` and `parsing` phases will be reported as Events.

All metadata available in `GraphQL` error are reported as attribute of the event:
 - `message`
 - `path`
 - `locations`
 - `code`: If it exists, the commonly used `error.extensions.code` allowing to group errors of the same kind.
 - `coordinate`: a new extension added by [this PR](https://github.com/ardatan/graphql-tools/pull/7588) allowing to identify the faulty resolver.

For the last attribute, the plugin have to override the execution function to enable a flag that adds schema coordinate to errors. This feature is only available when using our own fork of the graphql-js executor (which is the case by default).

# Side Quest

By modifying the batch executor, I also noted that the signal is just ignored when merging multiple execution requests.

I've fixed that by adding a new util to merge a list of signal into one that gets aborted only once all signals are aborted or GCed.

# Questions

This is changing how errors are reported in a few ways, which can break user dashboard or alerts depending on how they have written filters:
 - The name of the event is no longer `exception` but `graphql.error`
 - The event is attached to `graphql.operation` span instead of the specific phase span (`graphql.execute`, `graphql.validate` and `graphql.parse`)
 - The attributes names have changed. But they 

# TODO

 - [x] Handle signals in batched executor
 - [x] Migrate execution GraphQL errors to OTEL Event
 - [x] Migrate validation GraphQL errors to OTEL Event
 - [x] Migrate parsing GraphQL errors to OTEL Event
 - [x] Add tests


packages/batch-execute/src/mergeRequests.ts
```diff
@@ -18,6 +18,7 @@ import {
   VariableNode,
   visit,
 } from 'graphql';
+import { abortSignalAll } from '../../signal/src/abortSignalAll.js';
```
ardatan: Is this import correct?

PR #1642, opened 2025-10-29
enhance(runtime): support async return values in `progressiveOverride`

ardatan:
Support promises in `progressiveOverride` option

```ts
defineConfig({
    async progressiveOverride(label: string, context: GatewayContext) {
        if (label === 'my_label') {
            const serviceResponse = await fetch('http://example.com/should_override', {
                headers: {
                    'x-some-header': context.headers['x-some-header'],
                }
            });
            const result = await serviceResponse.json();
            return result?.override;
        }
        return false;
    }
})
```

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -54,6 +54,7 @@ jobs:
           - subscriptions-with-transforms
           - type-merging-batching
           - operation-field-permissions
+          - launchdarkly-override
           # TODO: support containers before enabling (we use nats image in e2e)
           # - event-driven-federated-subscriptions
     name: Convert ${{matrix.e2e}}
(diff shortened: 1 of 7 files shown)


.changeset/strange-wolves-knock.md
```diff
@@ -0,0 +1,24 @@
+---
+'@graphql-mesh/fusion-runtime': minor
+'@graphql-tools/federation': minor
+'@graphql-hive/gateway-runtime': minor
+---
+
+Support promises in `progressiveOverride` option
+
+```ts
+defineConfig({
```
enisdenjo: We should always think that our readers don't know how to configure the gateway.
```suggestion
import { defineConfig } from '@graphql-hive/gateway';
export const gatewayConfig = defineConfig({
```

PR #1653, opened 2025-10-30
feat(opentelemetry): do not trace prometheus requests by default

EmrysMyrddin:
# Description

This PR adds a default filter for HTTP spans to ignore prometheus metrics scraping requests.

Those request are not interesting to trace or monitor.

# Details

The logic can look a bit complex for such a simple option. I have to set this default filter only if the configuration doesn't already contain a filter function for http spans. But TypeScript makes it difficult because it is not happy about accessing field or doing spreading on boolean (which would actually work but hey, that would be too easy it suppose).

Related to GW-482

### DIFF
--- a/.changeset/late-parents-wash.md
+++ b/.changeset/late-parents-wash.md
@@ -1,0 +1,56 @@
+---
+'@graphql-hive/plugin-opentelemetry': minor
+'@graphql-mesh/plugin-prometheus': minor
+---
+
+New `ignoreRequest` OpenTelemetry API to allow other plugins (like Prometheus integration) to mark
+an HTTP Request to be excluded from OpenTelemetry tracing.
+
+```ts
+import { hive } from '@graphql-hive/gateway/opentelemetry/api';
+import { defineConfig } from 'graphql-hive/gateway';
+
+export const gatewayConfig = defineConfig({
+  openTelemetry: {
+    traces: true,
+  },
+  plugins: () => [
+    {
+      instrumentation: {
+        request: ({ request }) => {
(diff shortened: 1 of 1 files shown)


packages/plugins/opentelemetry/src/plugin.ts
```diff
@@ -959,3 +966,49 @@ function getURL(request: Request) {
 
   return new URL(request.url, 'http://localhost'); // to be iso with whatwg-node/server behavior
 }
+
+export const defaultHttpFilter: SpansConfig['http'] = ({ request }) => {
+  // Ignore Prometheus requests
+  if (request.url.endsWith('/metrics')) {
```
ardatan: This assumes that prometheus metrics are always at `/metrics`. But it is configurable in Prometheus plugin, no?
Maybe we can have a shared flag to decide if the request is prom request?

EmrysMyrddin: Humm difficult to do, because this will be executed before even the `onRequest` hook.

Or the prometheus plugin have to also implement `instrument.request` to mark the request (which is a possibility of course)

EmrysMyrddin: It's not very simple to do, because you end up again with plugin order issues ^^'
So it's more fragile in programmatic usage where we don't control the plugin order. And it also means that this little bit of code will not be inside the span.

EmrysMyrddin: I've refactored, let me know what you think :-)

ardatan: Looks better!

PR #1738, opened 2025-11-27
feat(gateway): add support for new Hive CDN mirror and circuit breaker

EmrysMyrddin:
## Description

Adds support for the new Hive CDN fetcher with the addition of a secondary mirror and a circuit breaker.

## Details

This PR also ensures the schema fetcher is actually disposed, since the circuit breaker has to be disposed to properly shut down the gateway.

Ref GW-535

## Todo

- [x] add mirror and circuit breaker options for schema fetcher
- [x] dispose fetcher on gateway disposal
- [x] add mirror and circuit breaker for Persisted Documents
- [x] use Hive Logger in all Hive related features

### DIFF
--- a/packages/runtime/tests/__snapshots__/hive.spec.ts.snap
+++ b/packages/runtime/tests/__snapshots__/hive.spec.ts.snap
@@ -1,41 +1,5 @@
 // Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
-exports[`Hive CDN respects env vars: hive-cdn 1`] = `
-"type Query {
-  foo: String
-}"
-`;
-
-exports[`Hive CDN should request using "/supergraph" pathname from cdn even if the user provides "" as the pathname: hive-cdn 1`] = `
-"type Query {
-  foo: String
-}"
-`;
-
-exports[`Hive CDN should request using "/supergraph" pathname from cdn even if the user provides "/artifacts/v1" as the pathname: hive-cdn 1`] = `
-"type Query {
-  foo: String
-}"
-`;
-
-exports[`Hive CDN should request using "/supergraph" pathname from cdn even if the user provides "/supergraph" as the pathname: hive-cdn 1`] = `
-"type Query {
-  foo: String
-}"
-`;
-
-exports[`Hive CDN should request using "/supergraph" pathname from cdn even if the user provides "/supergraph.graphql" as the pathname: hive-cdn 1`] = `
-"type Query {
-  foo: String
-}"
-`;
-
-exports[`Hive CDN should request using "/supergraph" pathname from cdn even if the user provides "/v1/artifacts" as the pathname: hive-cdn 1`] = `
-"type Query {
-  foo: String
-}"
-`;
-
 exports[`Hive CDN should request using "/supergraph(.graphql)" pathname from cdn even if the user provides "" as the pathname: hive-cdn 1`] = `
 "type Query {
   foo: String


packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -643,40 +671,51 @@ export function createGatewayRuntime<
...
-          logger: LegacyLogger.from(
-            configContext.log.child('[hiveSupergraphFetcher] '),
-          ),
+        const { endpoint, key, circuitBreaker } = config.supergraph;
+        const endpoints = (Array.isArray(endpoint) ? endpoint : [endpoint]).map(
+          (url) => (url.endsWith('/supergraph') ? url : `${url}/supergraph`),
+        );
 
+        const fetcher = createCDNArtifactFetcher({
+          endpoint: endpoints as [string, string],
```
ardatan: Maybe `asArray` can be used here instead of casting.

packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -383,22 +411,22 @@ export function createGatewayRuntime<
...
       'skipValidation' in config ? config.skipValidation : false;
-    const executorPlugin: GatewayPlugin = {
+    unifiedGraphPlugin = {
       onValidate({ params, setResult }) {
         if (shouldSkipValidation || !params.schema) {
           setResult([]);
         }
       },
-      onDispose() {
+      async onDispose() {
```
ardatan:
```suggestion
     onDispose() {
```

packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -383,22 +411,22 @@ export function createGatewayRuntime<
...
         pausePolling();
-        return transportExecutorStack.disposeAsync();
+        await transportExecutorStack.disposeAsync();
```
ardatan:
```suggestion
    return handleMaybePromise(() => transportExecutorStack.disposeAsync(), () => schemaFetcher.dispose?.());
```

EmrysMyrddin: I've used `async/await` here because `disposeAsync` type advertises that it always return a `Promise`, so I'm not sure `handleMaybePromise` is needed?

.changeset/cold-bags-confess.md
```diff
@@ -0,0 +1,67 @@
+---
+'@graphql-hive/gateway-runtime': minor
+---
+
+## New Hive CDN mirror and circuit breaker
```
enisdenjo: First line should be the name of the changeset, dont title it, it will look weird when listing all changesets.

package.json
```diff
@@ -63,6 +63,7 @@
     "vitest": "3.2.4"
   },
   "resolutions": {
+    "@graphql-hive/core": "0.16.0-alpha-20251126142740-d265dc6d6d73bf62ec597dbe41409f466b9d9874",
```
enisdenjo: Was this released? Do we want the alpha?

packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -643,40 +672,51 @@ export function createGatewayRuntime<
...
-        const fetcher = createSupergraphSDLFetcher({
-          endpoint,
-          key,
-          logger: LegacyLogger.from(
-            configContext.log.child('[hiveSupergraphFetcher] '),
-          ),
+        const { endpoint, key, circuitBreaker } = config.supergraph;
+        const endpoints = (Array.isArray(endpoint) ? endpoint : [endpoint]).map(
+          (url) => (url.endsWith('/supergraph') ? url : `${url}/supergraph`),
+        );
```
enisdenjo: Why are we doing this in the gateway? Shouldn't the supergraph fetcher handle this instead?

enisdenjo (changes requested):
A few small remarks, otherwise looking good.

packages/runtime/src/createGatewayRuntime.ts
EmrysMyrddin: Because It was handled by the schema fetcher before, not anymore (because there is no schema fetcher anymore, it's a generic artifacts fetcher)

EmrysMyrddin: So it's just to avoid a breaking change on our end

packages/runtime/src/createGatewayRuntime.ts
EmrysMyrddin: I can't, because `asArray` returns an array, not a tuple of 2 string

packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -280,35 +282,54 @@ export function createGatewayRuntime<
...
+            ? url
+            : joinUrl(
+                url.endsWith('/services')
+                  ? url.substring(0, url.length - 8)
+                  : url,
+                'sdl',
+              ),
+      );
+      const fetcher = createCDNArtifactFetcher({
+        endpoint: endpoints as [string, string],
```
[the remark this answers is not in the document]
enisdenjo: Fixed in upcoming commit.

PR #1796, opened 2025-12-15
⛑️ Fix potential runtime error in `useSubgraphExecuteDebug` plugin when `context.log` is undefined.

dimitriBouteille:
Fix potential runtime error in `useSubgraphExecuteDebug` plugin when `context.log` is undefined.

 **Problem**

  In `packages/runtime/src/plugins/useSubgraphExecuteDebug.ts:10`, the code uses optional chaining on `executionRequest.context` but not on the `log` property before calling `.child()`:

  ```typescript
  let log = executionRequest.context?.log.child('[useSubgraphExecuteDebug] ');
```

  This can cause a runtime error: Cannot read property 'child' of undefined if executionRequest.context exists but context.log is undefined or null.

  **Solution**

  Add optional chaining to the log property to safely handle cases where the logger is not available:

  ```typescript
  let log = executionRequest.context?.log?.child('[useSubgraphExecuteDebug] ');
```

  This change ensures the plugin gracefully handles missing loggers, which is consistent with the existing null checks on line 15 (if (!log || !shouldLog)).

### DIFF
--- a/packages/runtime/src/plugins/useSubgraphExecuteDebug.ts
+++ b/packages/runtime/src/plugins/useSubgraphExecuteDebug.ts
@@ -7,7 +7,7 @@ export function useSubgraphExecuteDebug<
 >(): GatewayPlugin<TContext> {
   return {
     onSubgraphExecute({ executionRequest }) {
-      let log = executionRequest.context?.log.child(
+      let log = executionRequest.context?.log?.child(
         '[useSubgraphExecuteDebug] ',
       );
       let shouldLog = false;


dimitriBouteille:
Hi @ardatan 

I use `@graphql-hive, graphql-yoga` and `next.js`, here is an example of my configuration:

@/api/mesh/createGateway

```js
import fs from 'fs/promises';
import path from 'path';
import { createGatewayRuntime } from '@graphql-hive/gateway-runtime';

export async function createMeshGateway() {
  const schemaPath = path.join(process.cwd(), '.mesh/schema.graphql');
  const supergraphSdl = await fs.readFile(schemaPath, 'utf-8');

  return createGatewayRuntime({
    supergraph: supergraphSdl,
  });
}
```


pages/api/graphql

```js
import { createYoga } from 'graphql-yoga';
import type { NextApiRequest, NextApiResponse } from 'next';

import { createMeshGateway } from '@/api/mesh/createGateway';

type YogaContext = {
  req: NextApiRequest;
  res: NextApiResponse;
};

// eslint-disable-next-line @typescript-eslint/no-explicit-any
let yoga: any = null;

async function getYoga() {
  if (yoga) {
    return yoga;
  }

  const gateway = await createMeshGateway();

  return createYoga<YogaContext>({
    graphqlEndpoint: '/api/graphql',
    schema: gateway.getSchema(),
  });
}

export default async function handler(
  req: NextApiRequest,
  res: NextApiResponse
) {
  const yogaInstance = await getYoga();
  return yogaInstance.handle(req, res);
}

export const config = {
  api: {
    bodyParser: {
      sizeLimit: '25mb',
    },
  },
};
```

ardatan:
The schema itself is not usable on its own. `GatewayRuntime` is already a Yoga instance so you should use that runtime directly instead of taking the schema from the gateway.

ardatan:
You can check this section for the local execution;
https://[the-guild.dev/graphql/mesh/v1/local-execution](https://the-guild.dev/graphql/mesh/v1/local-execution)
Or you can use the testing utility
https://the-guild.dev/graphql/hive/docs/gateway/other-features/testing/gateway-tester#quick-start

dimitriBouteille:
Perfect, it’s all good :)

Last question, I see that it is possible to pass a `context` option (in the `ExecutionRequest` type) there is documentation explaining what is possible to do with this context ?

ardatan:
Closing the PR then!
About your question, could you give an example of what you are trying to do?

PR #1838, opened 2026-01-07
feat: Support operation type timeouts

darren-west:
### Summary
Allow a factory function for the timeout. This supports the ability to not set timeouts on subscriptions which is often not required. 

For example:
```ts
     timeout: ({operationType}) => operationType !== 'subscription' ? 20000 : undefined,
```

### DIFF
--- a/packages/executors/http/tests/buildHTTPExecutor.test.ts
+++ b/packages/executors/http/tests/buildHTTPExecutor.test.ts
@@ -13,7 +13,7 @@ import {
 } from '@whatwg-node/server';
 import { GraphQLError, parse } from 'graphql';
 import { createSchema, createYoga } from 'graphql-yoga';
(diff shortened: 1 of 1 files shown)


packages/executors/http/src/index.ts
```diff
@@ -277,10 +277,14 @@ export function buildHTTPExecutor(
         return createResultForAbort(signalFromRequest.reason);
       }
       signals.push(signalFromRequest);
-    }
+
     if (options?.timeout) {
-      signals.push(AbortSignal.timeout(options.timeout));
+      const timeout = typeof options?.timeout === 'number' ? options.timeout : options?.timeout(request);
```
ardatan:
```suggestion
      const timeout = typeof options.timeout === 'number' ? options.timeout : options.timeout(request);
```

ardatan (approved):

PR #1855, opened 2026-01-14
Hive Console proxy schema fetcher uses fetch and identifies, Hive Console plugin's agent uses fetch

enisdenjo:


packages/runtime/tests/hive.spec.ts
```diff
@@ -137,7 +137,7 @@ describe('Hive CDN', () => {
               // @ts-expect-error - Fetch signature is not compatible
               return upstreamServer.fetch(url, opts);
             }
-            return gateway.fetchAPI.Response.error();
+            return fetch(url, opts);
```
ardatan, line 140: Why do we need that?

enisdenjo: onFetch is invoked when gateway is polling the proxy schema

enisdenjo: same goes for supergraph schema, see the changes in the cdn fetcher

ardatan: But this is the test. We mock the invalid response here instead of making an actual http request.

enisdenjo: no no, we need a valid response - from the disposable server started before. please check out the full test

ardatan: Oh my bad so it is a real HTTP server. So we used the real server because fetcher doesn't take our fetch previously.
Then we can actually drop the actual HTTP server, and use `fetch` from `createServerAdapter` of `cdnServer` directly.

packages/runtime/src/createGatewayRuntime.ts
```diff
@@ -305,9 +305,13 @@ export function createGatewayRuntime<
           ? // no endpoint.map just to make ts happy without casting
             [ensureSdl(endpoint[0]), ensureSdl(endpoint[1])]
           : ensureSdl(endpoint),
-        circuitBreaker,
         accessKey: key,
         logger: configContext.log.child('[hiveSchemaFetcher] '),
+        // @ts-expect-error - MeshFetch is not compatible with `typeof fetch`
+        fetch: configContext.fetch,
```
ardatan, line 311: Maybe we should have an eslint rule that prevents us using the globals by accident.

enisdenjo: I dont think it would help here because the `fetch` propery was completely omitted, so global fetch was not used in this code

ardatan: I meant in general we might use `fetch` directly by mistake like below;
```ts
function something({ fetch }) {
  fetch(..) // This is from the arguments
}
```

```ts
function something(refactorarg) {
  fetch(..) // not this is NOT from the arguments anymore but the globals
}
```

ardatan (approved):

PR #1932, opened 2026-02-06
feat: add mcp plugin implementation

adambenhassen:


examples/mcp-example/QUICK_START.md
```diff
@@ -0,0 +1,57 @@
...
+
+# Call a tool
+curl http://localhost:4000/mcp \
+  -H "Content-Type: application/json" \
+  -d '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"get_users","arguments":{"limit":5}}}'
+```
+
+## Resources
+
+- Documentation (WIP): https://docs.google.com/document/d/14FeFPz40QMMbNE2w-9OPzwOe8GcSEjJJQoKSKkm5xdY
```
enisdenjo: This doc is not publicly accesible.

examples/mcp-example/QUICK_START.md
```diff
@@ -0,0 +1,57 @@
...
+- Examples and config reference: https://github.com/graphql-hive/gateway/blob/7356319c23b005c29a44efc06885e936fbcb8923/examples/mcp-example/README.md#diff-e382d6a0e278f4d8f7da4a3d96e106221517ecb1d9d141d7b9d15014f05569bb
```
enisdenjo: "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

packages/plugins/mcp/src/description-provider.ts
```diff
@@ -0,0 +1,223 @@
+import type { ResolvedToolConfig } from './plugin.js';
+
+/** Configuration object for a description provider, identified by `type` with provider-specific fields. */
+export interface DescriptionProviderConfig {
+  /** Provider type identifier (e.g. "langfuse") */
+  type: string & {};
```
enisdenjo: `string & {}`?

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1523 @@
...
+  return {
+    onSchemaChange({ schema: newSchema }) {
+      try {
+        const newRegistry = new ToolRegistry(resolvedTools, newSchema);
+        const newOptions = buildHandlerOptions(newRegistry);
+        schema = newSchema;
+        registry = newRegistry;
+        mcpHandlerOptions = newOptions;
+      } catch (err) {
+        console.error(
```
enisdenjo: Consider using the gateway's logger, if not available in the hook - pass it as an argument to the plugin.

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1523 @@
...
+    }: {
+      request: Request;
+      url: URL;
+      setRequestParser: (
+        parser: (req: Request) => {
+          query: string;
+          variables?: Record<string, unknown>;
+        },
+      ) => void;
+    }) {
```
enisdenjo:
```suggestion
    }) {
```

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1523 @@
...
+      const ctx = mcpToolCalls.get(request);
+      if (!ctx) {
+        if (url.pathname === mcpPath) {
+          console.error(
```
enisdenjo: Use gateway's logger, there is one in `serverContext`.

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1523 @@
...
+      request: Request;
+      setResultProcessor: (
+        processor: (
+          result: unknown,
+          fetchAPI: unknown,
+          mediaType: string,
+        ) => Response | Promise<Response>,
+        mediaType: string,
+      ) => void;
+    }) {
```
enisdenjo:
```suggestion
    }) {
```

packages/plugins/mcp/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@graphql-hive/plugin-mcp",
+  "version": "0.0.1",
```
enisdenjo: We dont use zero based versioning in the GW. If it's a prerelease, it's an alpha/beta, otherwise its a v1.

```suggestion
  "version": "0.0.0",
```

packages/plugins/mcp/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "@graphql-hive/logger": "workspace:^",
+    "@graphql-mesh/fusion-runtime": "workspace:^",
+    "tslib": "^2.8.1"
+  },
+  "devDependencies": {
+    "@graphql-hive/gateway-runtime": "workspace:*",
+    "graphql": "^16.12.0",
+    "graphql-yoga": "^5.16.2",
+    "pkgroll": "2.23.0",
+    "sofa-api": "0.18.9-alpha-20260209105916-0e8bbdbf1143e58f25a464acfb9132491bff932a"
```
enisdenjo: I guess intentional alpha release? We of course dont want to release a stable depending on an alpha.

enisdenjo (changes requested):
- [Examples are auto-generated from e2e tests](https://github.com/graphql-hive/gateway/blob/main/examples/README.md), please create tests for the examples
- Use the gateway's hive logger instead of the console, I flagged a few places, but theres many more

packages/plugins/mcp/package.json
ardatan: Let's keep it 0.0.0 and let changesets bump it to v1 with a major changeset

enisdenjo: Right 👍  (updated suggestion)

packages/plugins/mcp/package.json
adambenhassen: Yes, it's intentional. the plugin uses SOFA's `resolveFieldType` and `buildSchemaObjectFromType` for gql-to-json schema conversion, but SOFA doesn't export these publicly yet. This alpha package refers to changes from https://github.com/graphql-hive/SOFA/pull/1852 that only expose these functions.

packages/plugins/mcp/package.json
adambenhassen: I agree. It's a byproduct of debugging the CI not running/publishing the packages

packages/plugins/mcp/src/description-provider.ts
adambenhassen: byproduct of debugging, will clean up

examples/mcp-example/QUICK_START.md
adambenhassen: It's private, will be unreferenced

packages/plugins/mcp/package.json
ardatan: https://github.com/graphql-hive/SOFA/pull/1852/changes#r2997340580
Let's add a changeset there then I'll merge and release so we can get rid of the alphas here

adambenhassen: Done.

ardatan: Released

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1823 @@
...
+  const internalRequests = new WeakSet<Request>();
+
+  function docsToSource(docs: { body: string }[]): string {
+    return docs.map((d) => d.body).join('\n');
+  }
+
+  function startHiveInit() {
+    if (!hiveInitFailed && hiveInitPromise) return;
+    hiveInitFailed = false;
+    hiveInitPromise = hiveLoader!
```
n1ru4l: NIT: I consider the usage of `!` a code-smell

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1823 @@
...
+    `Operations path "${config.operationsPath}" is neither a file nor a directory`,
+  );
+}
+
+/**
+ * Create a Gateway plugin that exposes GraphQL operations as MCP tools.
+ * Handles the full MCP protocol (initialize, tools/list, tools/call, resources)
+ * by routing tool calls through the Yoga GraphQL pipeline.
+ */
+export function useMCP(ctx: PluginContext, config: MCPConfig): GatewayPlugin {
```
n1ru4l, line 889: Is there any hook used that is exclusive to the Gateway and does not exist in Yoga? If so, should this be typed as a Yoga plugin instead, to imply it can also be used with any GraphQL server?

packages/plugins/mcp/README.md
```diff
@@ -0,0 +1,497 @@
...
+    mimeType: 'image/png',
+    blob: 'iVBOR...',
+  },
+  { name: 'schema', uri: 'files://schema.graphql', file: './schema.graphql' },
+];
+```
+
+Agents discover resources via `resources/list` and fetch them with `resources/read`. Templates allow dynamic URIs with parameters.
+
+## Hive integration
```
n1ru4l: This feature introduces a runtime dependency of the Hive GraphQL API for production traffic critical APIs. This should have been mentioned and flagged in the design phase. 

APIs that are essential for our customers core business to work should always poll from the CDN endpoints ([behind a circuit breaker](https://the-guild.dev/graphql/hive/docs/schema-registry/high-availability-resilience)) to ensure high availability and reduce the amount of failure points.

I studied the functionality and I think we could easily built this on top of the Hive CDN by introducing two minor changes:

## 1. Publish an app deployment manifest to the CDN upon app publish

During the app deployment creation we we publish a manifest that is available via `https://cdn.graphql-hive.com/artifacts/v1/<target_id>/apps/my-app/1.0.0`.

That manifest contains information such as:

```json5
{
  "isActive": true,
  "name": "my-app",
  "version": "1.0.0",
  "documents": [
    "hash1",
    "hash2",
    "hash3",
    // ....
  ]
}
```

That way the plugin can discover all available documents based on the CDN.

## 2. Tag an app deployment

Instead of relying paginating the the Hive GraphQL API for finding the latest app deployment, we can introduce a tagging or pinning functionality similar to how docker tags (e.g. `latest`) works.

That could mean that the manifest of a version that is tagged is also published to another shorthand URL. e.g. `https://cdn.graphql-hive.com/artifacts/v1/<target_id>/apps/my-app/tags/latest`.

```bash
hive app:tag <appName>@<appVersion> <tagName>
```

_____

What do you think? Happy to have some follow up discussion, help in further speccing out the proposed changes on Console side and assist in finding implementing a solution.

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1823 @@
...
+/** Defines how a tool's GraphQL operation is sourced, either inline or by reference to a named operation. */
+export interface MCPToolSource {
+  /** Source type: 'inline' for a query string, 'graphql' for a named operation reference */
+  type: string;
+  /** The GraphQL operation source (required when type is 'inline') */
+  query?: string;
+  /** Name of the operation to resolve (required when type is 'graphql') */
+  operationName?: string;
+  /** Whether the operation is a query or mutation (required when type is 'graphql') */
+  operationType?: string;
```
n1ru4l: Is there any specific reason for not using a discriminating union for distinguishing the `type` and enforce whether the `query` parameter is optional or not on a TypeScript level? This would get rid of code-smells such as using `!`.

Example:
```
type InlineMCPToolSource = {
  type: 'inline';
  query: string;
  operationName?: string;
  operationType?: string;
}

type GraphQLMCPToolSource = {
  type: 'inline';
  operationName: string;
  operationType: string;
}

export type MCPToolSource = InlineMCPToolSource | GraphQLMCPToolSource
```

packages/plugins/mcp/examples/demos/01-basic-gateway.ts
```diff
@@ -0,0 +1,117 @@
+// Step 1: Plain GraphQL gateway, no MCP yet.
+//
+// curl -s http://localhost:4000/graphql -H "Content-Type: application/json" -d '{"query":"{ weather(location: \"Tokyo\") { temperature conditions humidity } }"}' | jq
+
+import { createServer } from 'node:http';
```
ardatan: Maybe we can have these tests as E2E tests, and then generate example based on those using our example generator. What do you think @enisdenjo ?

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1823 @@
...
+
+    async onRequest({
+      request,
+      url,
+      endResponse,
+      requestHandler,
+      serverContext,
+      fetchAPI,
+    }) {
+      // Block external GraphQL access when disableGraphQLEndpoint is set
```
ardatan: I'm not fan of this recursion of the request handling to be honest. If unhandled route handler is the blocker, it can be bypassed by setting `graphqlEndpoint` to something like `/{graphql, mcp}`.

packages/plugins/mcp/src/hive-loader.ts
```diff
@@ -0,0 +1,396 @@
...
+    projectSlug: parts[1]!.trim(),
+    targetSlug: parts[2]!.trim(),
+  };
+}
+
+const REQUEST_TIMEOUT_MS = 30_000;
+const PAGE_SIZE = 100;
+
+const FETCH_DOCS_QUERY = `
+  query FetchAppDeploymentDocs($reference: TargetReferenceInput!, $appName: String!, $appVersion: String!, $first: Int!, $after: String) {
```
ardatan: This means we have to maintain the Hive Console API usage in our repository. Could it exist in Hive Console SDK then we can use here just like Persisted Documents fetcher, supergraph fetcher etc?

packages/plugins/mcp/src/schema-converter.ts
```diff
@@ -0,0 +1,245 @@
...
+  type GraphQLOutputType,
+  type GraphQLSchema,
+  type GraphQLType,
+  type SelectionSetNode,
+} from 'graphql';
+import { buildSchemaObjectFromType, resolveFieldType } from 'sofa-api';
+
+export { resolveFieldType, buildSchemaObjectFromType };
+
+/**
```
ardatan, line 21: Nit: Can we extract the graphql operation to json schema logic in a seperate package so we can reuse it in SOFA if SOFA's didn't work here?

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1823 @@
...
+      `Cannot access operations path "${config.operationsPath}": ${err instanceof Error ? err.message : String(err)}`,
+    );
+  }
+
+  if (stat.isFile()) {
+    return readFileSync(opsPath, 'utf-8');
+  }
+
+  if (stat.isDirectory()) {
+    const files = readdirSync(opsPath).filter((f) => f.endsWith('.graphql'));
```
ardatan: Do you think we can use this instead of maintaing an extra loader?
https://the-guild.dev/graphql/tools/docs/documents-loading

n1ru4l (changes requested):
Thanks for the work on this!

After reviewing implementation details of the Hive Console integration, I don't think we should move forward with this approach in its current form. The main concern is the introduction of a runtime dependency on the Hive GraphQL API for production-critical traffic. For APIs that are essential to customers' core business, we need to minimize points of failure and prioritize high availability.

I proposed an alternative approach that requires some further work on Hive Console side, but that effort is achievable in a reasonable time frame. Happy to help and collaborate on that matter to get this to the finish line! In the future I would recommend to have the the implementation plan reviewed with the team ahead of time, so such concerns can be raised earlier.

There are also some smaller points around typescript strictness and portability (Gateway vs Yoga), but to me those are secondary compared to the architectural constraint above.

I did not spend excessive amount on reviewing the core functionality of the actual MCP tool server (whether it provides the correct results according to the MCP spec), but I have trust that it covers the requirements of our customers based on the initial design document (added link it in the description; dunno if something newer or another iteration exists right now).

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1823 @@
...
+    [key: string]: DescriptionProvider | Record<string, unknown> | undefined;
+  };
+  /** Suppress outputSchema from all tools in tools/list responses */
+  suppressOutputSchema?: boolean;
+  /** Block direct access to the GraphQL endpoint (only MCP path is accessible) */
+  disableGraphQLEndpoint?: boolean;
+  /** Hive App Deployment as an operation source. Fetches persisted documents; those with @mcpTool directives are auto-registered as tools, and all documents are available as named operations for tools with source.type: 'graphql'. */
+  hive?: MCPHiveConfig;
+  /** Maximum request body size in bytes for the MCP endpoint (default: 1048576 = 1MB) */
+  maxRequestBodySize?: number;
```
ardatan: Hmm handling the max body size of the request is interesting, because I think it is more like HTTP server level thing, no?

packages/plugins/mcp/src/hive-loader.ts
adambenhassen: There's a plan to migrate this to a CDN-based approach, which would eliminate the need for direct API usage and a maintained query. Until those CDN endpoints exist on the Console side, this query is the working implementation and for the short-term, it is sufficient to provide a full app deployment functionality

packages/plugins/mcp/src/schema-converter.ts
adambenhassen: Yes, I've already noted your suggestion. Let's work on that in the next cycle, it's not a blocker.

packages/plugins/mcp/examples/demos/01-basic-gateway.ts
adambenhassen: These scenarios are already covered in the project's E2E tests (`e2e/mcp-plugin/mcp-plugin.e2e.ts`) and goes even further. The demos are standalone runnable examples for documentation purposes.

Here's what's already covered by the e2e tests:
```
1. Initialize (protocol version, capabilities, server info)
2. tools/list (all configured tools returned)
3. tools/list schemas (inputSchema and outputSchema)
4. tools/call inline (executes and returns structured data)
5. tools/call unknown tool (protocol error)
6. Mutation tool (with/without confirmation)
7. File based tool (operationName)
8. @mcpTool directive (auto registers from operations file)
9. Description provider (resolves tool description)
10. Per field description provider (overrides input field description)
11. Argument alias (renames input, de-aliases on call)
12. output.path (extracts subset of response data)
13. Postprocess hook (transforms result to text)
14. Preprocess hook (short circuits execution)
15. @mcpDescription on variable (resolves from provider)
16. @mcpDescription on selection field (resolves from provider)
17. Resources (list, read text, read blob, unknown URI error)
18. Resource templates (list and resolve with handler)
19. Header forwarding (MCP to internal dispatch)
20. Tool annotations (destructiveHint, readOnlyHint)
21. Content annotations (audience, priority in response)
```

ardatan: Great but we have E2E tests and generate examples from them so that we make sure that the examples are valid and tested with the latest packages. Are those examples tested somehow in somewhere? If not, then this means we will need to maintain those examples without testing.

adambenhassen: I see what you mean now, that's a good point! Maybe I could add a test to execute those?

packages/plugins/mcp/src/plugin.ts
adambenhassen: I've looked into this and noticed that this has big implications, so from what I've gathered so far:
- `loadDocuments` from `@graphql-tools/load` is async only, no sync variant
- No async startup hook in the `GatewayPlugin` lifecycle, so we'd need to defer loading to request time for local files
- `loadFilesSync` from `@graphql-tools/load-files` works but returns `DocumentNode[]`, not strings, which requires a `print()` then `parse()` round trip since `loadOperationsFromString` expects raw strings
- Could refactor `loadOperationsFromString` to accept `DocumentNode` directly to avoid the round trip, but that's a bigger change

I agree with avoiding an extra loader, so let's tackle this in a follow up PR

ardatan: > loadDocuments from @graphql-tools/load is async only, no sync variant

There is sync variant of it `loadDocumentsSync` -> [graphql-tools/packages/load/src/documents.ts at master · ardatan/graphql-tools](https://github.com/ardatan/graphql-tools/blob/master/packages/load/src/documents.ts#L51)

> Could refactor loadOperationsFromString to accept DocumentNode directly to avoid the round trip, but that's a bigger change

Isn't it just a one-line of code change to remove this? https://github.com/graphql-hive/gateway/pull/1932/changes?mode=single#diff-50997fe375aed04b014e7e951b224ad682370049ecb2c78c34b74dc809f64131R230
And if you expect a single DocumentNode not `DocumentNode[]` then you can do `concatAST(documentNodes)` to get a single `DocumentNode` from multiple ones.

packages/plugins/mcp/examples/demos/01-basic-gateway.ts
ardatan: Ideally each example should be an individual E2E test instead of a single E2E test directory covering all.

packages/plugins/mcp/src/plugin.ts
adambenhassen: I understand your reasoning. I initially had it implemented without the recursion as you previously mentioned but the reason I had it route through `requestHandler` is for observability, when MCP requests go through as `/graphql`, tracing and monitoring tools capture them all under the generic `/graphql` path instead of `/mcp`, making MCP traffic invisible in metrics and traces. Using `/{graphql,mcp}` as the endpoint would bring back that regression since Yoga would treat both paths equivalently. Beyond tracing, preserving the `/mcp` pathname makes it straightforward for middlewares to identify MCP requests via URL pathname checking, rather than relying on custom headers like `x-mcp-internal` which I had added as a workaround for a customer. 

Maybe you know of a way to handle this while preserving `/mcp` path visibility in tracing and middleware?

ardatan: >  Using /{graphql,mcp} as the endpoint would bring back that regression since Yoga would treat both paths equivalently.

Unless you cut the request off with a custom request parser by detecting the endpoint. I didn't mean we should use `/graphql` for MCP requests. After setting `/{graphql, mcp}` endpoint, you can detect the MCP request by checking the pathname in `onRequestParse` (request.url is still there), then set your request parser when the pathname matches. If it is not set, it will use the default parser and Yoga will treat it as a regular GraphQL request.

ardatan: I have a PoC (Sofa V2 proposal) as I previously sent on Slack; https://guild-oss.slack.com/archives/CSFK5GGAU/p1772206139711299
Here I set the request parser based on the pathname dynamically;
https://stackblitz.com/edit/stackblitz-starters-igyrny?file=src%2Flib%2FgraphqlToJsonSchema.ts%3AL84

adambenhassen: Ok, that makes sense for the request parsing side. The concern I have is around `graphqlEndpoint` -- since it's set at server creation, the user would need to include all MCP paths in it. If they have multiple MCP endpoints, they'd need to maintain something like `/{graphql,mcp,mcpSearch,mcpAdmin}` and keep it in sync. Is there a way to mutate `graphqlEndpoint` after server creation, or a way for plugins to extend the accepted paths?

ardatan: It can be fully dynamic with "*" or a URL pattern like "/:pathparam". That graphqlEndpoint is to detect the graphql endpoint, and return the landing page in case of 404. I think it should be fine to set the graphql endpoint. Or we can even allow plugins to change it like you said. But all those would be better than the recursive request handling.

packages/plugins/mcp/src/plugin.ts
adambenhassen: Not a one-line change but I got your point. I honestly don't see much value from this change, it adds extra dependencies and some unnecessary changes to what is basically a simple mechanism to load operations from a file. But none-the-less, I've made the changes to end this madness.

packages/plugins/mcp/src/plugin.ts
adambenhassen: I think allowing plugins to extend the accepted paths would be the cleanest approach, that way the MCP plugin can register its paths without the user needing to configure `graphqlEndpoint` manually, and we avoid accepting everything with `*`

Would you be able to handle that on the Yoga side? Once that's in place I can refactor the MCP plugin to drop the recursion.

packages/plugins/mcp/examples/demos/01-basic-gateway.ts
enisdenjo: Yeah, I agree with Arda. There is no other place in code that has examples set up like this except for here. I'd prefer to have them be e2e tests. 

In general, code that does not connect (except for typechecking) tends to be forgotten and outdated realquick - so having example code as tests puts pressure and makes sure everyting is up-to-date.

Alternatively, move the examples to another repo and use it as an example-only repo.

packages/plugins/mcp/examples/README.md
```diff
@@ -0,0 +1,221 @@
...
+const myProvider: DescriptionProvider = {
+  async fetchDescription(toolName, config) {
+    return fetch(`https://my-cms.com/api/${config.prompt}`).then((r) =>
+      r.text(),
+    );
+  },
+};
+
+useMCP(ctx, {
+  ...yamlConfig,
```
enisdenjo: Where does this yamlConfig come from?

packages/plugins/mcp/package.json
```diff
@@ -58,7 +58,7 @@
     "@langfuse/client": "^5.0.2",
     "@opentelemetry/api": "^1.9.0",
     "graphql": "^16.13.2",
-    "graphql-yoga": "^5.19.0",
+    "graphql-yoga": "5.20.0-alpha-20260403171057-74b8645742b9e38cd0ec99c771f84f67d66ad535",
```
adambenhassen: TODO: Replace alpha build after https://github.com/graphql-hive/graphql-yoga/pull/4428 is merged
NOTE: This is also causing CI failures since CI resolves `graphql-yoga` from the root workspace.

enisdenjo (commented):
You have done a lot of work! I, and I think others, would appreciate a walkthrough - extensive description, video or a short meeting just to show us around in order to understand better how pieces connect - or maybe you have some other idea?

packages/plugins/mcp/examples/README.md
adambenhassen: `yamlConfig` is intentionally left abstract here. this snippet just shows the pattern for mixing YAML config with a custom provider. The actual YAML config is shown in the section above (`mcp.yaml` config). But sure, I'll add a comment to make that clearer.

packages/plugins/mcp/examples/demos/01-basic-gateway.ts
adambenhassen: As I mentioned earlier, all of these demo scenarios are already covered by the e2e tests. I initially didn't plan to include standalone demos in the PR, but Uri requested them. Happy to remove them.

packages/plugins/mcp/README.md
adambenhassen: Will include this during the planning of changes for console, thanks for the input

packages/plugins/mcp/src/plugin.ts
adambenhassen: There's a gateway logger coupling `(Logger.child()` from `@graphql-hive/logger`) that prevents this from being a standalone Yoga plugin. I initially thought `onSchemaChange` was Gateway-specific but it's an Envelop hook. So aside from the logger, nothing prevents it from being a Yoga plugin in the future.

packages/plugins/mcp/src/plugin.ts
```diff
@@ -0,0 +1,1851 @@
...
+      request,
+      url,
+      endResponse,
+      requestHandler,
+      serverContext,
+      fetchAPI,
+    }) {
+      // Block external GraphQL access when disableGraphQLEndpoint is set
+      if (
+        config.disableGraphQLEndpoint &&
```
ardatan: We can have this behavior by removing `/graphql` endpoint from `graphqlEndpoint` right?

adambenhassen: Sure but, the complication is the schema bootstrap in proxy mode, if `/graphql` is removed from the endpoint, the one-time `{ __typename }` request we send to trigger lazy schema loading would 404. The bootstrap exists because the gateway loads its schema lazily on first request, and the MCP plugin needs the schema to resolve tools. Is there another way to trigger schema loading without routing through a GraphQL endpoint?

ardatan: In CLI mode, schema loading is done on bootstrap. There is "getSchema" method of GW instance (Yoga instance also) that would trigger that but it is a dirty hack.
Ideally in onRequestParse hook inside the request parser, the schema should be ready. You can maybe move the logic inside the request parser.

adambenhassen: > Ideally in onRequestParse hook inside the request parser, the schema should be ready

That works for `tools/call`, I confirmed the schema is ready by the time our request parser runs. The problem is `tools/list` and other methods. Even if we move them to `onRequestParse`, the request parser needs to return `{ query, variables }` for GraphQL execution, but `tools/list` doesn't execute GraphQL, it just returns the registry contents.

Also, I tried using `getSchema()` but it's not accessible from the plugin code, we'll have to either expose it through `onYogaInit` or add it to the `YogaServer` itself.

ardatan: Actually "yoga" in onYogaInit is the gateway instance even if it is not typed well. So it has "getSchema()" method there.

adambenhassen: It isn't. I checked the source, `onYogaInit` passes `this` (the `YogaServer` instance, line 439 of `server.ts`), not the server adapter `Proxy` that `createYoga()` returns. `getSchema` is added by `createGatewayRuntime` via `Object.defineProperties` on the `Proxy` (line 1006 of `createGatewayRuntime.ts`), after `createYoga()` has already returned. So the `YogaServer` instance in `onYogaInit` doesn't have `getSchema` on it.

Here's also a snippet to back this:
``` ts
import { createGatewayRuntime } from '@graphql-hive/gateway-runtime';
let yogaFromInit: any = null;
const gateway = createGatewayRuntime({
  logging: false,
  proxy: { endpoint: 'http://localhost:4000/graphql' },
  plugins: () => [
    {
      onYogaInit({ yoga }: { yoga: any }) {
        yogaFromInit = yoga;
        console.log('onYogaInit: constructor:', yoga.constructor?.name);
        console.log('onYogaInit: typeof getSchema:', typeof yoga.getSchema);
        console.log('onYogaInit: has getSchema:', 'getSchema' in yoga);
        console.log('onYogaInit: own props:', Object.getOwnPropertyNames(yoga).filter((k) => k.includes('Schema')));
      },
    },
  ],
});
console.log('\n--After createGatewayRuntime--');
console.log('gateway: typeof getSchema:', typeof (gateway as any).getSchema);
console.log('gateway === yogaFromInit:', gateway === yogaFromInit);
console.log('yogaFromInit: typeof getSchema:', typeof yogaFromInit?.getSchema);
process.exit(0);
```

ardatan: I haven't re-checked tbh. But anyways maybe short-circuiting inside the request parser can solve it like here;<br>graphql-hive/graphql-yoga#4430

adambenhassen: Good. With this addition to `graphql-yoga`, we can move the logic to `onRequestParse`. Merge it and push a new release.

ardatan: https://github.com/graphql-hive/graphql-yoga/pull/4431

enisdenjo (commented):
Regarding the failing E2E for Docker, you have to mount the missing directory into Docker for it to have access to the files. For example:

```ts
const { gateway } = createTenv(__dirname);
const gw = await gateway({
  runner: {
    docker: {
      volumes: [
        {
          host: './file.ts',
          container: '/gateway/file.ts',
        },
        {
          host: './folder',
          container: '/gateway/folder',
        },
      ],
    },
  },
});
```

About the `plugins/mcp/examples/`, I'd really prefer it being an E2E (for reasons I mentioned before), but it's not a blocker at the moment. Please do make a task/issue for future reference. It's important our examples are actively tested and therefore maintained.

One thing I need clarification for: why have the MCP be a builtin CLI feature? I think having it be an installable plugin until we have more than one confirmed user would be a better (and safer?) way to go, no reason to install more deps for a feature that we dont yet know would even be used much. We can bundle the plugin in Docker, I am fine with that though.

What about docs? They're coming in next?

Otherwise, great work! After fixing the E2Es and deciding if MCP should really be a builtin, it's an approved PR from me.

packages/plugins/mcp/package.json
```diff
@@ -0,0 +1,66 @@
...
+    }
+  },
+  "dependencies": {
+    "@graphql-hive/logger": "workspace:^",
+    "tslib": "^2.8.1"
+  },
+  "devDependencies": {
+    "@graphql-hive/gateway-runtime": "workspace:*",
+    "@graphql-tools/graphql-file-loader": "^8.1.12",
+    "@graphql-tools/load": "^8.1.8",
```
ardatan: Adding these into `devDependencies` would cause them to be bundled into the plugin package. Then since it is part of CLI, the bundled code will be part of the CLI too. Not a blocker or an urgent thing but it would be nice to have them as dependencies for the nested dependency stability, you know.
Other than that, the PR overall looks good to me. Nothing to add besides what @enisdenjo said. Let's make sure we have issues/tasks for the things planned later.

adambenhassen:
> Regarding the failing E2E for Docker, you have to mount the missing directory into Docker for it to have access to the files. For example:
> 
> ```ts
> const { gateway } = createTenv(__dirname);
> const gw = await gateway({
>   runner: {
>     docker: {
>       volumes: [
>         {
>           host: './file.ts',
>           container: '/gateway/file.ts',
>         },
>         {
>           host: './folder',
>           container: '/gateway/folder',
>         },
>       ],
>     },
>   },
> });
> ```

Oh, so that's why those tests are failing!!1

> About the `plugins/mcp/examples/`, I'd really prefer it being an E2E (for reasons I mentioned before), but it's not a blocker at the moment. Please do make a task/issue for future reference. It's important our examples are actively tested and therefore maintained.

Absolutely agree with you. I'll issue a task to track this.

> One thing I need clarification for: why have the MCP be a builtin CLI feature? I think having it be an installable plugin until we have more than one confirmed user would be a better (and safer?) way to go, no reason to install more deps for a feature that we dont yet know would even be used much. We can bundle the plugin in Docker, I am fine with that though.

I have no solid arguments to back the need for it to be builtin into the CLI, except for my early-dev attempt at maintaining consistency across the codebase. Let's keep it separated indeed.

> What about docs? They're coming in next?

Yep, that's still cooking.

> Otherwise, great work! After fixing the E2Es and deciding if MCP should really be a builtin, it's an approved PR from me.

🙏 🎉 🎉 🎉

packages/plugins/mcp/package.json
adambenhassen: Good insight! I've looked into the dependencies, and pulled some out of `devDepdencies` and removed cli ref. Let me know if there's anything else left, would also love to hear your input on the changeset

enisdenjo (commented):
It's good to go if you ask me. But I'd like to see the CI pass first? What happened to Windows?

adambenhassen:
> It's good to go if you ask me. But I'd like to see the CI pass first? What happened to Windows?

No idea, must have been some change from upstream.

Commit 994a2927aa16d16c96d1e0ebd0c4c9884a331662 from upstream fixes it.

packages/plugins/mcp/src/hive-loader.ts
n1ru4l, line 1: As discussed yesterday, let's remove the hive-loader from this PR and move it to a WIP pull request. Then let's release the first version of the mcp plugin 👍

n1ru4l (changes requested):
🟢 for merging after from me https://github.com/graphql-hive/gateway/pull/1932/changes#r3078211202 is addressed 👍

dotansimha (changes requested):
I fully agree with the takes mentioned here about the hive-loader component. 

The architecture behind Console's CDN was designed to ensure our main cluster (=Console GraphQL API) is never part of the hotpath of part of the production runtime of our customers. 
Using the GraphQL API directly to fetch runtime information from Console is not ideal and might lead to us dealing with our customers incidents as ours. 

Let's drop it for now, and rethink how this will be loaded from the CDN.

ardatan (approved):
Looks good to me after fixing the conflicts and a ✅ CI

n1ru4l (approved):
Legit! Let's get it in 🥳

enisdenjo (approved):

dotansimha (approved):

PR #2006, opened 2026-02-24
feat: use hive lab instead of graphiql

mskorokhodov:
Replace old graphiql lab with new hive laboratory


enisdenjo (changes requested):
Looking good! Please write a changeset documenting the new graphiql default.

enisdenjo (changes requested):
Please write up a changelog and also how users can opt out. This is _almost_ a breaking change.

.changeset/small-mangos-crash.md
```diff
@@ -0,0 +1,18 @@
...
+You can always switch back to GraphiQL by updating the config:
+
+import { defineConfig } from '@graphql-hive/gateway';
+import { renderGraphiQL } from 'graphql-yoga';
+
+export const gatewayConfig = defineConfig({
+
+  renderGraphiQL: renderGraphiQL,
+
+});
```
enisdenjo: this is a code snippet, we use ``` for code snippets

.changeset/small-mangos-crash.md
```diff
@@ -0,0 +1,18 @@
...
+---
+
+Hive Laboratory is now the default editor
+
+We’ve upgraded the default GraphQL experience in Hive Gateway by replacing GraphiQL with Hive Laboratory — a more powerful, editor-style interface built for modern workflows.
+
+You can always switch back to GraphiQL by updating the config:
+
+import { defineConfig } from '@graphql-hive/gateway';
+import { renderGraphiQL } from 'graphql-yoga';
```
enisdenjo: Installing yoga just to get the graphiql renderer is an overkill, especially for something that is potentially breaking. I'd much rather prefer to have an option that is true/false and can be set through an environment variable. Something like `renderLegacyGraphiQL: true/false` and `RENDER_LEGACY_GRAPHIQL`.

packages/gateway/package.json
```diff
@@ -100,6 +100,7 @@
     "@graphql-hive/plugin-aws-sigv4": "workspace:^",
     "@graphql-hive/plugin-opentelemetry": "workspace:^",
     "@graphql-hive/pubsub": "workspace:^",
+    "@graphql-hive/render-laboratory": "0.1.6",
```
enisdenjo: Like with all other dependencies, please use ranged versions.

enisdenjo (changes requested):

.changeset/small-mangos-crash.md
```diff
@@ -0,0 +1,29 @@
...
+
+We’ve upgraded the default GraphQL experience in Hive Gateway by replacing GraphiQL with Hive Laboratory — a more powerful, editor-style interface built for modern workflows.
+
+You can always switch back to GraphiQL by updating the config:
+
+```
+import { defineConfig } from '@graphql-hive/gateway';
+
+export const gatewayConfig = defineConfig({
+  renderLegacyLaboratory: true
```
ardatan: Why "legacyLaboratory" instead of GraphiQL? It is a bit confusing to call GraphiQL like that, no?

mskorokhodov: updated to GraphiQL

.changeset/small-mangos-crash.md
```diff
@@ -0,0 +1,29 @@
...
+'@graphql-hive/gateway': minor
+---
+
+Hive Laboratory is now the default editor
+
+We’ve upgraded the default GraphQL experience in Hive Gateway by replacing GraphiQL with Hive Laboratory — a more powerful, editor-style interface built for modern workflows.
+
+You can always switch back to GraphiQL by updating the config:
+
+```
```
enisdenjo:
```suggestion
```ts
```

.changeset/small-mangos-crash.md
```diff
@@ -0,0 +1,29 @@
...
+import { defineConfig } from '@graphql-hive/gateway';
+
+export const gatewayConfig = defineConfig({
+  renderLegacyGraphiQL: true
+});
+```
+
+Or via CLI option
+
+```
```
enisdenjo:
```suggestion
```sh
```

.changeset/small-mangos-crash.md
```diff
@@ -0,0 +1,29 @@
...
+hive-gateway --render-legacy-graphiql
+```
+
+Or via env variable
+
+```
```
enisdenjo:
```suggestion
```ts
```

.changeset/small-mangos-crash.md
```diff
@@ -0,0 +1,29 @@
...
+RENDER_LEGACY_GRAPHIQL=TRUE hive-gateway
```
enisdenjo: why uppercase?

```suggestion
RENDER_LEGACY_GRAPHIQL=true hive-gateway
```

or 

```suggestion
RENDER_LEGACY_GRAPHIQL=1 hive-gateway
```

enisdenjo (commented):
note how the `optionEnv:jit` is set up, there's a comment there explainin why we do that for boolean env vars.  please do the same for legacy graphiql, I think it'd be something like:

```ts
.on('optionEnv:render-legacy-graphiql', function (this: Command) {
  this.setOptionValueWithSource(
    'renderLegacyGraphiQL',
    getEnvBool('RENDER_LEGACY_GRAPHIQL'),
    'env',
  );
})
```

otherwise, looks good. thanks! once you handle these few nits, rebase and fix the CI - we're good to go

.changeset/small-mangos-crash.md
ardatan: Is `legacy` accurate here? @enisdenjo What do you think?

enisdenjo: I think so yes. I wouldnt call it `deprecated` because it's perfectly valid to use it, we're just moving away from it.

ardatan: I didn't mean `deprecated` but `legacy` sounds like `graphiql`'s `legacy` version but GraphiQL is still an actively maintained alternative. So I couldn't make sure if it is misleading. Maybe it is too nitpicking

enisdenjo (approved):

PR #2031, opened 2026-02-27
Adding breaking test case for missing __typename on federated non abstract types

nopor:
The test case covers the missing addition of the `__typename` for non abstract types.
When a federated non abstract type is requested with `... on SomeType`, it's not expanded with `__typename`.
This has been the case in previous releases (e.g. @graphql-tools/delegate@10.0.16).

What behaviour is expected here, @ardatan?

### DIFF
--- a/packages/delegate/tests/finalizeGatewayRequest.test.ts
+++ b/packages/delegate/tests/finalizeGatewayRequest.test.ts
@@ -145,6 +145,48 @@ describe('finalizeGatewayRequest', () => {
     expect(print(filteredQuery.document)).toBe(print(expected));
   });
 
+  test('should expect __typename on non abstract types', () => {
+    const query = parse(/* GraphQL */ `
+      query bookingQuery($id: ID!) {
+        bookingById(id: $id) {
+          id
+          customer {
+            ... on Customer {
+              name
+            }
+          }
+        }
+      }
+    `);
+    const filteredQuery = finalizeGatewayRequest(
+      {
+        document: query,
+        variables: {
+          id: 'b1',
+        },
+      },
+      {
+        targetSchema: bookingSchema,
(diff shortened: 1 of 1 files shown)


nopor:
Right.
The delegated requests are missing the `__typename` for federated types.
The schema would look like this:
```
type Query {
  getSomething: Something!
}
type Something {
  id: ID!
  name: String!
  federatedType: FederatedType!
}
type FederatedType {
  id: ID!
  name: String!
}
```
Service A would satisfy `Something` whereas `FederatedType` would be satisfied by Service B.

The requests currently look like this:
Service A receives:
```
{"query":"query($representations:[_Any!]!){_entities(representations:$representations){__typename ...on Something{federatedType{...on FederatedType{id}}id}}}","variables":{"representations":[{"id":"some-id","__typename":"Something"}]}}
```

Service B receives:
```
{"query":"query($representations:[_Any!]!){_entities(representations:$representations){__typename ...on FederatedType{id name}}}","variables":{"representations":[{"id":"federated-id"}]}}
```

This then fails in Service B with the following error:
```
"errors": [
    {
      "message": "Missing \"__typename\" argument",
      "path": [
        "getSomething",
        "federatedType"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "message": "Missing \"__typename\" argument",
          "path": [
            "getSomething",
            "federatedType"
          ]
        },
        "gateway": "0.0.0"
      }
    }
```

The expected request for Service B would be:
```
{"query":"query($representations:[_Any!]!){_entities(representations:$representations){__typename ...on FederatedType{id name}}}","variables":{"representations":[{"id":"federated-id","__typename":"FederatedType"}]}}
```

PR #2243, opened 2026-04-12
fix(delegate): prevent redundant type merging

qsona:
The fallback name-based matching in resolveExternalValue matched unrelated subschemas when neither had an explicit name set, because `undefined === undefined` evaluated to true. This caused type merging to always trigger even when fetching directly from the source subschema, resulting in a redundant second call.

Add a null guard to skip the fallback when subschema names are not set.

Regression introduced in a54b0c17 (#1557). I have verified that the newly added test fails at a54b0c17 and passes at the preceding commit, 058ef2f8 .

Please feel free to adjust the new test if it doesn't match the project's style.

### DIFF
--- a/.changeset/fix-redundant-type-merging.md
+++ b/.changeset/fix-redundant-type-merging.md
@@ -1,0 +1,7 @@
+---
+'@graphql-tools/delegate': patch
+---
+
+Fix redundant type merging calls when subschemas have no explicit name.
+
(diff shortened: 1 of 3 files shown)


ardatan:
Could you add a changeset that explains the issue and the fix? Thanks!

qsona:
@ardatan I've added a changeset. Thank you.

ardatan:
Thanks @qsona !

PR #2305, opened 2026-04-30
Fix Laboratory migration

ardatan:
Followup https://github.com/graphql-hive/gateway/pull/2006

- Add E2E tests to test behavior of `renderLegacyGraphiQL` flag for both proxy and supergraph modes
- Add `proxy` mode support to E2E testkit
- Fix incorrectly handled `renderLegacyGraphiQL` option in the CLI. `attributeName` is not working well because it is not reflected to the TS types. So it is better to handle it where we prepare the configuration for the runtime.
- Remove `renderLegacyGraphiQL` option from the runtime because it has nothing to do with the runtime. Laboratory is only added to CLI.
- `playgroundName` option is added to the runtime so `GraphiQL` reference will be `Laboratory` when the laboratory is used. And GraphiQL Name won't incorrectly be used
- Landing page wasn't built when `yarn build` is called but only built before release which looks unsafe. So fixed it. Now on every `yarn build`, landing page will be rebuilt.


ardatan:
@copilot apply changes based on the comments in [this thread](https://github.com/graphql-hive/gateway/pull/2305#pullrequestreview-4207291552)

PR #2318, opened 2026-05-05
Add requestDeadline option for hard end-to-end request time limits

enisdenjo:
The existing `requestTimeout` only covers the time to receive the request from the client - once the body is fully received, the timer is cancelled and the handler runs indefinitely. This leaves no way to enforce a hard cap on the total request lifecycle.

This PR adds a new `requestDeadline` option to `ServerConfig` that sets a hard end-to-end time limit in milliseconds from connection to response completion. When the deadline is exceeded, the server responds with a 503 and closes the connection.

Documented in https://github.com/graphql-hive/docs/pull/105.


packages/gateway/src/servers/bun.ts
```diff
@@ -75,9 +79,29 @@ export async function startBunServer<TContext extends Record<string, any>>(
...
+            resolve(
+              new Response('Request deadline exceeded', {
+                status: 503,
+                headers: { Connection: 'close' },
+              }),
+            ),
+          );
+        }),
+      ]);
+    }
```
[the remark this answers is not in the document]
enisdenjo: - there is no way to cancel an abortsignal.timeout
- if we remove the onAbort in .finally of the race - the second promise will never resolve
- I added an onAbort check

e2e/subgraph-timeout/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "@e2e/subgraph-timeout",
+  "private": true
+}
```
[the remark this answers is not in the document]
enisdenjo: no need

packages/gateway/src/servers/nodeHttp.ts
```diff
@@ -79,6 +85,24 @@ export async function startNodeHttpServer<TContext extends Record<string, any>>(
     );
   }
 
+  if (requestDeadline) {
+    server.on('request', (req: IncomingMessage, res: ServerResponse) => {
+      const timer = setTimeout(() => {
+        if (!res.headersSent) {
+          res.writeHead(503, { Connection: 'close' });
```
ardatan, line 92: Isn't 504 more accurate?

packages/gateway/src/servers/nodeHttp.ts
```diff
@@ -79,6 +85,24 @@ export async function startNodeHttpServer<TContext extends Record<string, any>>(
...
+          res.end('Request deadline exceeded', () => {
+            // wait for the response to flush before destroying the socket
+            req.socket?.destroy();
+          });
+        } else {
+          req.socket?.destroy();
+        }
+      }, requestDeadline);
+      res.on('finish', () => clearTimeout(timer));
```
ardatan:
```suggestion
      res.once('finish', () => clearTimeout(timer));
```

packages/gateway/src/servers/nodeHttp.ts
```diff
@@ -79,6 +85,24 @@ export async function startNodeHttpServer<TContext extends Record<string, any>>(
...
+      res.on('close', () => clearTimeout(timer));
```
ardatan:
```suggestion
      res.once('close', () => clearTimeout(timer));
```

packages/gateway/src/servers/nodeHttp.ts
```diff
@@ -79,6 +85,20 @@ export async function startNodeHttpServer<TContext extends Record<string, any>>(
...
+      const timer = setTimeout(() => {
+        if (!res.headersSent) {
+          res.writeHead(503, { Connection: 'close' });
+          res.end('Request deadline exceeded');
+        }
+        req.socket?.destroy();
+      }, requestDeadline);
+      res.on('finish', () => clearTimeout(timer));
+      res.on('close', () => clearTimeout(timer));
+    });
```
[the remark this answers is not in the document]
enisdenjo: thats ok

packages/gateway/src/servers/nodeHttp.ts
enisdenjo: 504 gateway timeout is better suited for `upstreamTimeout`s where the upstream caused the gateway to time out. this is not the case, the gateway can time out a request without the upstream being an issue

packages/gateway/src/servers/bun.ts
ardatan, line 85: Maybe this helps?
https://bun.com/reference/bun/Server/timeout

enisdenjo: no, that's the `idleTimeout` override - it measures silence between packets, not total elapsed time - a slow but steady client will never trigger it.

packages/gateway/src/servers/bun.ts
```diff
@@ -75,9 +79,29 @@ export async function startBunServer<TContext extends Record<string, any>>(
...
+          deadlineSignal.addEventListener('abort', () =>
+            resolve(
+              new Response('Request deadline exceeded', {
+                status: 503,
+                headers: { Connection: 'close' },
+              }),
+            ),
+          );
+        }),
+      ]);
```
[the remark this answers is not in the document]
enisdenjo: we'll live with it, fix is too ugly. left a comment

ardatan (approved):

PR #2382, opened 2026-05-25
fix(delegate): preserve inline fragment type condition during cross-subgraph delegation

### LINKED ISSUE #2381: Inline Fragment Type Condition Lost During Schema Stitching Delegation - Causing field selection to be missing from Subgraph Request
## Summary

When delegating fields across subgraphs, the `extractUnavailableFieldsFromSelectionSet` function in `@graphql-tools/delegate` strips the inline fragment type condition (`... on ConcreteType`) when the source subgraph's schema doesn't register the concrete type as a subtype of the parent interface. This causes the delegated request to place fields directly on the interface type, where they don't exist, resulting in the fields being silently removed from the outgoing request.

**Affected package**: `@graphql-tools/delegate`  
**Function**: `extractUnavailableFieldsFromSelectionSet`  
**Impact**: Fields inside inline fragments on interface types are completely omitted from the delegated request to the target subgraph, so the data is never fetched.

---

### Setup

Given this schema across two subgraphs (simplified names):

```graphql
extend schema @link(
  url: "https://specs.apollo.dev/federation/v2.3"
  import: ["@key", "@inaccessible", "@shareable", "@override", "@extends", "@interfaceObject", "@external", "@provides", "@requires"]
)

type Query {
  relByAcct: Relationship
}

interface P {
  id: ID
}

type Account
@key (fields: "id", resolvable: false) @interfaceObject {
    id: ID
}

type Relationship implements P @key(fields: "id") {
  id: ID
  relType: String
  acct: Account @shareable
}
```

```graphql
extend schema @link(
  url: "https://specs.apollo.dev/federation/v2.3"
  import: ["@key", "@inaccessible", "@shareable", "@override", "@extends", "@interfaceObject", "@external", "@provides", "@requires"]
)

interface Account @key(fields: "id") {
  id: ID
}

type Relationship
@key(fields: "id") {
  id: ID
  acct: Account @shareable
}

type Card implements Account @key(fields: "id") {
  id: ID
  billing: String
}

```

Note: **Subgraph A** (the first schema) defines `acct` as returning an `Account` interface but does NOT define the concrete implementing type `Card`. `Card` only exists in Subgraph B's schema.

### Query

```graphql
query {
  relByAcct {
    acct {
      ... on Card {
        billing
      }
    }
    id
  }
}
```

### Expected Behavior

1. Subgraph A resolves `relByAcct` and `id`.
2. `billing` is unavailable in Subgraph A → delegated to Subgraph B.
3. Subgraph B receives an entity resolution request that still contains the inline fragment type condition — e.g.:
   ```graphql
   query {
     _entities(representations: [{ __typename: "Relationship", id: "cr-1" }]) {
       ... on Relationship {
         acct {
           ... on Card {
             billing
           }
         }
       }
     }
   }
   ```
4. Subgraph B resolves `billing` on `Card` ✅

### Actual Behavior (Bug)

Internally, the gateway constructs the selection set for the entity query but drops the inline fragment wrapper:

```graphql
acct {
  billing
}
```

**The `... on Card` wrapper is missing.** `billing` is placed directly on the `Account` interface, where it doesn't exist. The `filterSelectionSet` function strips it before the request is sent.

Subgraph B actually receives:

```graphql
query {
  _entities(representations: [{ __typename: "Relationship", id: "cr-1" }]) {
    ... on Relationship {
      acct {
        __typename
      }
    }
  }
}
```

`billing` is **completely absent** from the delegated request — the data is never fetched.

---

## Root Cause

In `@graphql-tools/delegate/dist/index.cjs`, the function `extractUnavailableFieldsFromSelectionSet` handles inline fragments at ~line 2665:

```js
} else if (selection.kind === graphql.Kind.INLINE_FRAGMENT) {
  const subFieldName = selection.typeCondition?.name.value || fieldType.name;  // "Card"
  const subFieldType = selection.typeCondition && schema.getType(subFieldName) || fieldType;

  if (subFieldName === fieldType.name ||
      (isObjectType(subFieldType) || isInterfaceType(subFieldType)) &&
      isAbstractType(fieldType) &&
      schema.isSubType(fieldType, subFieldType)) {
    // BRANCH A: Card IS a known subtype of Account in this schema
    // → Inline fragment is preserved ✅
    const unavailableFields = extractUnavailableFieldsFromSelectionSet(...);
    if (unavailableFields.length) {
      unavailableSelections.push({
        ...selection,  // preserves ... on Card
        selectionSet: { selections: unavailableFields }
      });
    }
  } else if (isObjectType(subFieldType) || isInterfaceType(subFieldType)) {
    // BRANCH B: Card is NOT a known subtype (e.g., defined in another subgraph)
    // → Inline fragment is LOST ❌
    for (const subSelection of selection.selectionSet.selections) {
      if (shouldAdd(subFieldType, subSelection)) {
        unavailableSelections.push(subSelection);  // ← BARE field, no wrapper!
      }
    }
  }
}
```

**Branch B** fires when:

- The source subgraph (A) defines `Account` as an interface
- But does NOT have `Card` as an implementing type in its schema
- `schema.isSubType(Account, Card)` returns `false`

In this branch, the inline fragment's selections are pushed **without** the `... on Card` wrapper. The type condition is discarded.

### What happens next

1. `getFieldsNotInSubschema` wraps the result into the `acct` field node:

   ```graphql
   acct { billing }  ← no type condition
   ```

2. The delegation plan routes `acct` to Subgraph B

3. `filterSelectionSet` validates the request against Subgraph B's schema:
    - Visits `billing` as a field on `Account` (the interface)
    - `typeInfo.getFieldDef()` returns `null` (interface doesn't define `billing`)
    - **Field is stripped** (returns `null` from the visitor)

4. `acct` ends up with an empty selection set → also removed

5. Subgraph B receives a request with no meaningful selections → returns nothing useful

gonzo-32:
Closes: https://github.com/graphql-hive/gateway/issues/2381

## The Fix

Preserve the inline fragment wrapper:

```js
} else if (graphql.isObjectType(subFieldType) || graphql.isInterfaceType(subFieldType)) {
  const unavailableSubSelections = [];
  for (const subSelection of selection.selectionSet.selections) {
    if (subSelection.kind === graphql.Kind.FIELD && subSelection.name.value === "__typename") {
      continue;
    }
    if (shouldAdd(subFieldType, subSelection)) {
      unavailableSubSelections.push(subSelection);
    }
  }
  if (unavailableSubSelections.length) {
    // Preserve the inline fragment wrapper so the type condition is not lost
    unavailableSelections.push({
      ...selection,
      selectionSet: {
        kind: graphql.Kind.SELECTION_SET,
        selections: unavailableSubSelections
      }
    });
  }
}
```

### What changed

| Before (bug)                                                         | After (fix)                                                                                                          |
| -------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- |
| `unavailableSelections.push(subSelection)` — pushes bare field nodes | `unavailableSelections.push({ ...selection, selectionSet: { selections } })` — wraps in the original inline fragment |
| Delegated query: `acct { billing }`                                   | Delegated query: `acct { ... on Card { billing } }`                                                                   |
| `filterSelectionSet` strips `billing` (not on interface)              | `filterSelectionSet` finds `billing` on `Card` ✅                                                                     |

---

### DIFF
--- a/.changeset/delegate-inline-fragment-context.md
+++ b/.changeset/delegate-inline-fragment-context.md
@@ -1,0 +1,7 @@
+---
+'@graphql-tools/delegate': patch
+---
+
+Fix cross-subgraph delegation for unavailable fields selected through inline fragments or fragment spreads on interface fields.
+
+The fix in `extractUnavailableFieldsFromSelectionSet` within `extractUnavailableFields.ts` now preserves the fragment wrapper and type condition, so concrete-type fields do not lose their type context and get treated as missing.
--- a/packages/delegate/src/extractUnavailableFields.ts
+++ b/packages/delegate/src/extractUnavailableFields.ts
@@ -119,6 +119,7 @@ export function extractUnavailableFieldsFromSelectionSet(
           });
         }
       } else if (isObjectType(subFieldType) || isInterfaceType(subFieldType)) {
+        const unavailableSubSelections: SelectionNode[] = [];
         const subFieldTypeFields = subFieldType.getFields();
         for (const subSelection of selection.selectionSet.selections) {
           if (subSelection.kind === Kind.FIELD) {
@@ -130,10 +131,20 @@ export function extractUnavailableFieldsFromSelectionSet(
               subFieldTypeFields[subSelection.name.value];
             if (!subSelectionField) {
               if (shouldAdd(subFieldType, subSelection)) {
-                unavailableSelections.push(subSelection);
+                unavailableSubSelections.push(subSelection);
               }
             }
           }
+        }
+        if (unavailableSubSelections.length) {
+          // Preserve the inline fragment wrapper so the type condition is not lost
+          unavailableSelections.push({
+            ...selection,
+            selectionSet: {
+              kind: Kind.SELECTION_SET,
+              selections: unavailableSubSelections,
+            },
+          });
         }
       }
     } else if (selection.kind === Kind.FRAGMENT_SPREAD) {
--- a/packages/delegate/tests/extractUnavailableFields.test.ts
+++ b/packages/delegate/tests/extractUnavailableFields.test.ts
@@ -1,6 +1,7 @@
 import { makeExecutableSchema } from '@graphql-tools/schema';
 import { parseSelectionSet } from '@graphql-tools/utils';
 import {
+  FragmentDefinitionNode,
   getOperationAST,
   isObjectType,
   Kind,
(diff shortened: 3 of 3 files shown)


ardatan:
Thanks @gonzo-32 !

PR #2401, opened 2026-06-03
Default `mergeDirectives` to true and fix `false` by stripping directive usages

n1ru4l:
Custom directive definitions from subschemas were silently dropped from the stitched schema unless `mergeDirectives: true` was explicitly passed to `stitchSchemas`. This meant a schema like:

```graphql
directive @public on SCHEMA | OBJECT | FIELD_DEFINITION

type Query @public {
  isAnExample: Boolean @public
}
```

would stitch into a broken schema where `@public` was used on types and fields but never defined - because directive usages were carried over via AST nodes through `mergeCandidates`, but the directive definitions themselves were not collected.

The old `mergeDirectives: false` behavior was also broken for the inverse reason: it only dropped directive definitions, but left all the usages on field and type AST nodes intact. So it never achieved a clean "no directives" schema - it just produced an inconsistent one. The correct suppression would have required also stripping usages from every field, type, input field, and enum value AST node, which was never done.

This PR fixes both sides:

- `mergeDirectives` now defaults to `true`, so directive definitions are always collected from subschemas and retained in the stitched schema. This is the expected (now fixed) default behavior - a stitched schema should reflect the full set of directives its subschemas declare.
- `mergeDirectives: false` now does the full job: directive definitions are not collected, and all custom directive usages are stripped from type, field, input field, and enum value AST nodes. Built-in directives (`@deprecated`, `@specifiedBy`, etc.) are preserved regardless.

### DIFF
--- a/.changeset/purple-monkeys-drum.md
+++ b/.changeset/purple-monkeys-drum.md
@@ -1,0 +1,29 @@
+---
+'@graphql-tools/stitch': patch
+---
+
+Always retain directive definitions from subschemas in stitched schema
+
+Custom directive definitions from subschemas were silently dropped from the stitched schema unless `mergeDirectives: true` was explicitly passed to `stitchSchemas`. This meant a schema like:
+
+```graphql
+directive @public on SCHEMA | OBJECT | FIELD_DEFINITION
+
+type Query @public {
+  isAnExample: Boolean @public
+}
+```
+
+would stitch into:
+
+```graphql
+type Query @public {
+  isAnExample: Boolean @public
+}
+```
+
+The directive usages on types and fields were preserved (carried over via AST nodes through `mergeCandidates`), but the directive definitions themselves were dropped, producing a broken schema where directives are used but never defined.
+
+`mergeDirectives` was flawed to begin with... `mergeDirectives: false` produced an internally inconsistent schema where directive usages existed on fields and types but the corresponding definitions were removed. The correct way to suppress directives from appearing in the stitched schema would have been to also strip their usages from field and type AST nodes - but that was never done. So `mergeDirectives: false` never achieved a clean "no directives" result, it just produced a broken one. In essence, fixing the `mergeDirectives: false` behavior would actually be a breaking change!
+
+Having said all that, we've instead removed the `mergeDirectives` argument and guard so directives from subschemas are always collected. This is correct default behavior - a stitched schema should retain all directive definitions from its subschemas unconditionally. This is not breaking because it actually fixes the partial merge.
--- a/.changeset/purple-monkeys-drum.md
+++ b/.changeset/purple-monkeys-drum.md
@@ -2,7 +2,7 @@
 '@graphql-tools/stitch': patch
 ---
 
-Always retain directive definitions from subschemas in stitched schema
+Fix `mergeDirectives` option and default it to `true`
 
 Custom directive definitions from subschemas were silently dropped from the stitched schema unless `mergeDirectives: true` was explicitly passed to `stitchSchemas`. This meant a schema like:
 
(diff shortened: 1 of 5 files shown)


packages/stitch/tests/stitchSchemas.test.ts
```diff
@@ -3846,3 +3847,33 @@ it('should be able to extend a transformed schema', async () => {
...
+      ?.astNode?.directives?.find((d) => d.name.value === 'public'),
+  ).toBeDefined();
+  expect(
+    stitchedSchema
+      .getQueryType()
+      ?.getFields()
+      ['isAnExample']?.astNode?.directives?.find(
+        (d) => d.name.value === 'public',
+      ),
+  );
```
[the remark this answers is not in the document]
enisdenjo: -.-

dotansimha (approved):
Yes. This feel very much like a bug. Denis, I like the take on removing this field as it's just adding ambiguity on top of the incorrect behavior.

.changeset/purple-monkeys-drum.md
```diff
@@ -0,0 +1,29 @@
+---
+'@graphql-tools/stitch': patch
+---
```
n1ru4l, line 3: Removing an existing option mergeDirectives is definitely a breaking change to me.

Without having looked into the code I would assume that `mergeDirectives: false` would result in the first occurrence of the directive within the subgraphs to win and be kept in the stitched SDL. `@foo(a: String) on WHATNOT` and `@foo(b: String) on WHATEVER` would result in `@foo(a: String) on WHATNOT`
`mergeDirectives: true` would result in all definitions to be merged e.g. `@foo(a: String) on WHATNOT` and `@foo(b: String) on WHATEVER` would result in `@foo(a: String, b: String) on WHATNOT | WHATEVER`

n1ru4l (commented):
Since this change removes configuration from the API surface this is definitely a breaking change to me. Aside from that I am fine with shipping this if versioned as such.

enisdenjo:
> Since this change removes configuration from the API surface this is definitely a breaking change to me. Aside from that I am fine with shipping this if versioned as such.

`buildTypeCandidates` is an internal function not exposed to the users. The other function that is exposed has the property marked as depricated and not removed.

.changeset/purple-monkeys-drum.md
enisdenjo: mergeDirectives: true

```
@foo(a: String) on WHATNOT 
+
@foo(b: String) on WHATEVER
=
@foo(a: String, b: String) on WHATNOT | WHATEVER
```

mergeDirectives: false

```
@foo(a: String) on WHATNOT 
+
@foo(b: String) on WHATEVER
=
<nothing>
```

Have you read the PR description?

> [...] fixing the `mergeDirectives: false` behavior would actually be a breaking change!

This is because if we were to fix `mergeDirectives: false` then we'd have to remove all directive usages from the resulting schema.

So your test would actually merge into

```graphql
schema {
  query: Query
}

type Query {
  isAnExample: Boolean
}
```

and not 

```graphql
schema {
  query: Query
}

type Query @public {
  isAnExample: Boolean @public
}
```

Maybe you're trying to say that the former is less breaking?

n1ru4l:
the descriptions says, so does the changeset :D 
> we have removed the `mergeDirectives` argument

But did not specify where, I assumed issa from the stitchSchemas function! It is nowhere mentioned that the argument is deprecated and passing it is now ignored

.changeset/purple-monkeys-drum.md
n1ru4l: >This is because if we were to fix mergeDirectives: false then we'd have to remove all directive usages from the resulting schema.

It depends kind of on what that configuration option should have done in the first place. I just communicated what I would have expected sorely based on the config option name. If my assumption is wrong that is fine with me.

n1ru4l: That leads me to another question: Why not make `mergeDirectives` work as you lined out in a major release and set it to the default value `true`?

enisdenjo (approved):

enisdenjo (approved):

dotansimha (approved):

PR #2449, opened 2026-06-26
feat(runtime): graceful supergraph reload via generation overlap

advoretsky:
Closes #2448

On reload the previous generation (executor + subgraph transports) was disposed immediately, aborting any in-flight operation on it with a SCHEMA_RELOAD/503 error. Queries were retried by useRetryOnSchemaReload; mutations were not, so a mutation in flight at a reload failed.

Add opt-in graceful reload ("generation overlap"): keep a superseded generation alive and route only new requests to the new one. Operations are reference-counted per generation for their whole lifetime (across all subgraph hops) via onExecute/onExecuteDone, so in-flight queries and mutations finish on the schema they were admitted on. A superseded generation is disposed once idle, or force-disposed after a configurable drain timeout; the number of concurrently live generations is capped. Subscriptions are not overlapped and still end on reload to reconnect. On shutdown every live generation is disposed.

Configured via gracefulSchemaReload: { drainTimeout, maxConcurrentGenerations }; disabled by default (previous immediate-abort behavior preserved).

PR #2466, opened 2026-07-09
[DO NOT MERGE] feat(runtime): graceful supergraph reload via generation overlap

enisdenjo:
Squashed #2449 for alpha release.


enisdenjo:
#2449 is being merged

PR #2467, opened 2026-07-13
fix(router-runtime): BatchFetch drops $representations for empty entity alias, failing sibling fetches

vyacheslav-voloshyn:
## Problem

When the query planner emits a `BatchFetch` node that groups multiple entity fetches for the **same subgraph** into one aliased `_entities` request, and **one alias resolves to zero representations** (e.g. a nullable federated reference that is `null` for the given parent), the batched request breaks.

`buildBatchFetchVariables` (`packages/router-runtime/src/executor.ts`) skips attaching the `$representations` variable for the empty alias, but the generated `_entities` document still **declares** it as a required `[_Any!]!`. The outgoing operation therefore declares a variable it never provides, so the subgraph rejects the **entire** operation (`Variable "$..." of required type "[_Any!]!" was not provided`) — which also fails the sibling aliases that *did* have representations. Every entity in that batch comes back `null`.

## Reproduction

Two entity types resolved by the **same** subgraph, requested in one operation, where one reference is nullable:

```graphql
type Parent {
  optionalChild: Child          # nullable, resolved by subgraph B
  requiredChildren: [Child!]!   # resolved by subgraph B
}
```

Query both `optionalChild { ... }` and `requiredChildren { ... }` for a `Parent` whose `optionalChild` is `null`. The planner batches both `Child` fetches into a single `_entities` call. At runtime the `optionalChild` alias has zero representations, so its `$representations_*` variable is not sent — but the document still declares it — and subgraph B rejects the whole request, so `requiredChildren` resolve to `null` too.

## Fix

Always attach the representations variable, using `[]` when it's empty. `_entities(representations: [])` is valid and returns `[]`, so the sibling aliases hydrate normally.

```diff
   for (const [variableName, state] of batchContext.representationsByVariableName.entries()) {
-    if (!state.representations.length) {
-      continue;
-    }
     const targetVariables = variablesForFetch ?? (Object.create(null) as Record<string, any>);
     if (!variablesForFetch) variablesForFetch = targetVariables;
     targetVariables[variableName] = state.representations;
   }
```

Reproduced through `@graphql-hive/router-runtime@1.4.14`.

## Notes

- A changeset is included (`patch` bump for `@graphql-hive/router-runtime`).
- Happy to add a regression test — a pointer to the preferred harness for exercising `BatchFetch` execution against mock subgraphs would help me place it correctly.

### DIFF
--- a/.changeset/batchfetch-empty-representations.md
+++ b/.changeset/batchfetch-empty-representations.md
@@ -1,0 +1,9 @@
+---
+"@graphql-hive/router-runtime": patch
+---
+
+Fix `BatchFetch` execution dropping the `$representations` variable for an entity alias that resolved to zero representations.
+
+When a `BatchFetch` plan node groups several entity fetches for the same subgraph into one aliased `_entities` request and one alias resolves to an empty representation list (e.g. a nullable federated field that is `null`), `buildBatchFetchVariables` omitted that alias's variable. The batched document still declares it as a required `[_Any!]!`, so the operation declared a variable it never provided and the subgraph rejected the whole request — failing the sibling aliases that *did* have representations.
+
+Empty aliases now send `[]`, keeping the operation valid (`_entities(representations: [])` simply returns `[]`).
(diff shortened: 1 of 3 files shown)


enisdenjo (changes requested):
thank you for the great PR! would you mind craeting a regression test to make sure we got this coevered for the future?

enisdenjo (approved):
thanks! much appreciated

PR #2484, opened 2026-07-22
Reduce peak memory consumption for stitchSchemas

jdolle:
A minor improvement to memory usage. This places several of the larger objects inside utility functions and/or blocks to inform the runtime that it can free these objects in memory earlier.

For these high memory consumption functions, this is the difference between running out of memory and being able to process a larger schema.

A further improvement would be if we could gradually append new schemas onto a stitched schema, so that the older GraphQLSchema object could be freed. But this would require much more refactoring.

### DIFF
--- a/.changeset/late-tigers-sin.md
+++ b/.changeset/late-tigers-sin.md
@@ -1,0 +1,5 @@
+---
+'@graphql-tools/stitch': patch
+---
+
+Reduce peak memory consumption for stitchSchemas function
--- a/.changeset/late-tigers-sin.md
+++ b/.changeset/late-tigers-sin.md
@@ -2,4 +2,4 @@
 '@graphql-tools/stitch': patch
 ---
 
-Reduce peak memory consumption for stitchSchemas function
+Reduce peak memory consumption for `stitchSchemas` function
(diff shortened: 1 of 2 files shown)


packages/stitch/src/stitchSchemas.ts
```diff
@@ -50,99 +51,58 @@ import {
...
-    extensionASTNodes: schemaDefs.schemaExtensions,
-    extensions: null,
-    assumeValid: rest.assumeValid,
-  });
-
-  for (const extension of extensions) {
-    schema = extendSchema(schema, extension, {
+  if (extensions.length > 0) {
+    const mergedExtensions = concatAST(extensions);
+    schema = extendSchema(schema, mergedExtensions, {
```
jdolle, line 101: This refactor prevents calling extendSchema on each extension. Instead, it calls it once. This is important because extendSchema performs a deep copy of the graphqlschema. If there are multiple extensions then this places a lot of unnecessary work on the garbage collector

.changeset/late-tigers-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-tools/stitch': patch
+---
+
+Reduce peak memory consumption for stitchSchemas function
```
n1ru4l:
```suggestion
Reduce peak memory consumption for `stitchSchemas` function
```

n1ru4l (approved):

enisdenjo (approved):

PR #2485, opened 2026-07-22
fix(executor-http): cancel the response body when an SSE subscription ends

adam-drag:
## Problem

`handleEventStreamResponse` calls `reader.releaseLock()` on normal teardown:

```ts
stop
  .then(() => {
    subscriptionCtrl?.abort();
    if (body.locked) {
      reader.releaseLock();      // detaches the reader; body is never cancelled
    }
  })
  .catch((err) => {
    reader.cancel(err);          // cancel() only on the rejection path
  });
```

`releaseLock()` releases the reader's lock on the stream, but it does not cancel the stream. The response body stays active, so the underlying HTTP connection is never released and the socket remains `ESTABLISHED` until the process exits.

Note the asymmetry: the error path already uses `cancel()`. Only the normal-teardown path does not.

## Impact

A consumer that opens SSE subscriptions and later disposes them accumulates one leaked socket per disposed subscription. In our case a GraphQL gateway reached **~10,000 concurrent connections against fewer than 200 requests/hour**, and the upstream service eventually OOM'd. The connection count dropped to zero only on process restart.

Below, before and after patching
image

## Fix

```diff
-      reader.releaseLock();
+      reader.cancel().catch(() => {});
```

`.catch(() => {})` because cancelling an already-errored stream can reject, and teardown should not surface that.

## Evidence

Measured against a real upstream with per-socket tracking - recording the exact local ports opened by a test client, then re-checking those specific ports after killing it. Builds were stamped with the package version and a per-run nonce so there was no ambiguity about which code actually executed.

| Build | Sockets still open after the client dies |
|---|---|
| `3.3.0` (stock) | 0 of 12 closed |
| `3.3.0` + this change | **12 of 12 closed within 5s** |

A stock control run in the same session leaked 12/12, so the one line is the difference.

## Notes for reviewers

**No unit test accompanies this.** I tried to add one to `packages/executors/http/tests/handleEventStreamResponse.test.ts`, but the existing harness can't reach the changed code path: on a synthetic `ReadableStream` that never enqueues, `iterator.return()` never resolves and `stop` never settles, so `stop.then()` never runs. The only way I found to make the assertion fire was to call `readableStream.cancel()` from the test itself - which sets the observable directly and makes the test vacuous. I'd rather ship no test than one that passes regardless of the fix. If there's an idiomatic way to drive that path here, I'm happy to add it.

That behaviour may be worth a separate look on its own: `return()` alone does not appear to unpark a stalled reader. In the real-world path the request's abort signal is what settles `stop`, which is why this fix takes effect there.

**Two test files already fail on a clean `main`** - `packages/runtime/tests/subgraph-errors.test.ts` and `packages/runtime/tests/subscriptions.test.ts` (5 failures). I verified they're unrelated by stashing this change and re-running: identical `5 failed | 3 passed` both ways. Flagging it because one of them is a *subscriptions* test and could easily be misattributed to this PR.

## Checks run locally

On Node v26.5.0 (per `.node-version`) with yarn 4.13.0:

| Job | Result |
|---|---|
| `check:format` | pass |
| `check:lint` | pass |
| `check:types` | pass |
| `check:missing-peer-deps` | pass |
| `build` | pass |
| `bundle` | pass |
| `test` (unit) | 1704 passed, 5 pre-existing failures noted above |

Changeset included (`patch` bump for `@graphql-tools/executor-http`).


enisdenjo (commented):
this is looking great, we appreciate your work! it would be awesome if you could add in a test that makes sure we dont regress - I think it's a great test to have regardless. could you please handle that? 🙏

adam-drag:
> this is looking great, we appreciate your work! it would be awesome if you could add in a test that makes sure we dont regress - I think it's a great test to have regardless. could you please handle that? 🙏

@enisdenjo yeah sure, I agree, definitely worth adding a test. I will do that later this week. Cheers

enisdenjo (approved):
brilliant, thank you for the contribution!

enisdenjo:
@adam-drag seems like the leak tests are failing now? could be something missing during cleanup? could you take a look please

adam-drag:
@enisdenjo should be good now

Issue #2487, opened 2026-07-22
delegate drops nested variable values under graphql-js v17


sgohlke:
## Summary

Under **graphql-js v17**, `@graphql-tools/delegate` silently drops the **values** of every
variable that feeds a **nested (non-root) field argument** when it builds a delegated request.
The delegated document still *declares* and *references* those variables, but they are absent
from the forwarded variable values, so the downstream subschema receives `null` for them and
applies its defaults.

Root-field argument variables are unaffected (their values come from the already-coerced
resolver `args`), which is why the bug is easy to miss: only *nested* arguments break.

The trigger is the graphql-js v16 -> v17 change to the runtime shape of `info.variableValues`.
delegate still reads it as a flat `{ [name]: value }` map; in v17 the runtime values moved
under `info.variableValues.coerced`.

## Affected versions

| Package | Version |
|---|---|
| `graphql` | 17.0.2 (any 17.x) |
| `@graphql-tools/delegate` | 12.0.19 (latest published at time of writing) |
| `@graphql-tools/stitch` | 10.1.24 |
| `@graphql-tools/utils` | 11.2.0 |

`@graphql-tools/delegate` declares `graphql: "^14.0.0 || ^15.0.0 || ^16.0.0 || ^17.0.0"` as a
peer dependency, i.e. v17 is claimed as supported, but nested variable delegation is broken on it.
Works correctly on graphql 16.x. There is no fixed upstream release to upgrade to; the latest
published versions still reproduce it.

## Minimal reproduction

```ts
import { stitchSchemas } from '@graphql-tools/stitch'
import { stitchingDirectives } from '@graphql-tools/stitching-directives'
import type { Executor } from '@graphql-tools/utils'
import { buildSchema, graphql, print } from 'graphql'

const captured: { query: string; variables: unknown }[] = []
const executor = (({ document, variables }): unknown => {
    captured.push({ query: print(document), variables })
    return { data: { parent: { id: 'p1', items: [{ id: 'i1', children: [{ id: 'c1' }] }] } } }
}) as Executor

const { stitchingDirectivesTransformer } = stitchingDirectives()
const gateway = stitchSchemas({
    subschemaConfigTransforms: [stitchingDirectivesTransformer],
    subschemas: [{
        executor,
        schema: buildSchema(`
            type Child { id: ID! }
            type Item { id: ID!, children(offset: Int, limit: Int): [Child!]! }
            type Parent { id: ID!, items: [Item!]! }
[...]

PR #2544, opened 2026-08-23
add support for coerced values for graphql-17

egoodwinx:
fix for https://github.com/graphql-hive/gateway/issues/2487

create new getCoercedVariableValues function that supports graphql-js 16 and 17 and then use this variable to map the problem functions


enisdenjo (changes requested):
this looks good, can you please add a changeset so we can release this? thanks!

enisdenjo (approved):

PR #2545, opened 2026-08-24
fix: non-nullable batch execution

### LINKED ISSUE #1288: Cannot return null for non-nullable field
Let's say we have a subgraph that looks like this:

```graphql
type Query {
   foo: Foo!
}
```

and when we send the following query to the gateway:

```graphql
query {
   foo1: foo
   foo2: foo
}
```

Now let's say the aliased query `foo1` fails with an error, the subgraph will respond with something like:

```json
{
 "errors": [
    {
      "message": "Lorem ipsum.",
      "locations": [...],
      "path": ["_v0_foo"],
      "extensions": [...]
    }
  ],
  "data": null
}
```

We'd expect this response to be correctly returned from the gateway, however the graphql validation fails and the gateway throws this error:

```
Cannot return null for non-nullable field Query.foo.
```

We understand that due to the fact that `foo` is non nullable field partial responses are not allowed in this case, however since the root object data is returned as null we'd expect the response to be correctly returned in this case.

We're using hive as nestjs module:

```
@graphql-hive/nestjs: 1.0.15
graphql: 16.11.0
```

Are we missing something or is this actually a bug?

omahili:
Fixes #1288. 

Per the GraphQL spec [here](https://spec.graphql.org/October2021/#sec-Handling-Field-Errors) a non-nullable field returning `null` should propagate to the parent, in this case it should propagate to `data`. Batched sub-requests share one merged document so `null` should propagate to every result and not just the one whose fields actually failed.

I added a couple of tests for both single and multiple executions. Let me know if this is ok or if you have a better idea on the best way forward.


enisdenjo (changes requested):
this is looking great, thank you for contributing! would you mind adding a changeset so we can release this? simply run `yarn changeset` and follow the instructions, thanks!

omahili:
Happy to hear that, thank you! Added a changeset.

enisdenjo (approved):

PR #2554, opened 2026-09-01
feat(gateway): spread WebSocket closes across a shutdown drain window

m-sanders:
Stacked on #2553, whose commit is included in this diff.

A clean `1001` doesn't stop a rolling deploy from handing back every subscription at once. graphql-ws' `shouldRetry` defaults to `isLikeCloseEvent`, so it retries `1001` exactly like a dropped socket, and its default first `retryWait` is 1.3-4s.

Contract:

- New config option `websocketDrainTimeout`, in ms, default `0`. Config-only, like `gracefulShutdownTimeout`.
- `0` closes every client at once, as today. Set it and clients are closed in batches roughly a second apart across the window: `round(window / 1000)` batches capped at the client count, `ceil(clients / batches)` per batch, `floor(window / batches)` apart.
- `gracefulShutdownTimeout` still bounds the last batch's handshake, so a shutdown can take as long as both.
- Node only. Bun runs `startBunServer`, not `nodeHttp.ts`, and its shutdown drops WebSocket clients rather than closing them - so there is nothing for a drain window to spread, and the e2e test that asserts `1001` is skipped there. See the Bun section on #2553.

The name is the obvious bikeshed — happy to rename, or to derive the drain from `gracefulShutdownTimeout` instead of adding an option.


enisdenjo (changes requested):
there seems to be some CI tests failing, can you check them out please?

PR #2591, opened 2026-09-02
fix(stitch): do not mutate the caller's resolvers in stitchSchemas

### LINKED ISSUE #2592: stitchSchemas mutates the caller's resolvers, leaking a schema generation per supergraph reload
`stitchSchemas` mutates the resolver objects it is given. `addLocalFieldResolvers` wraps every user resolver whose field returns a merged type so the merged type's fields get resolved, and writes the wrapper back into the caller's field config:

```ts
const baseResolve = originalResolve ?? defaultMergedResolver;
const wrappedResolve = (parent, args, context, info) =>
  handleMaybePromise(
    () => baseResolve(parent, args, context, info),
    (result) => resolveLocalFieldResult(result, context, info, stitchingInfo, providedFields),
  );
if (existing != null && typeof existing === 'object') {
  existing.resolve = wrappedResolve;   // the caller's object
}
```

`wrappedResolve` closes over `baseResolve`, which is whatever `existing.resolve` was when the call started, and over this call's `stitchingInfo`. Pass the same `resolvers` object to `stitchSchemas` twice and the second wrapper wraps the first; pass it N times and you get a chain of N wrappers, each one holding a different `stitchingInfo`.

Observed on `@graphql-tools/stitch` 10.2.2; 10.2.3 has the same code (`packages/stitch/src/stitchSchemas.ts`).

### Why this is a problem

`@graphql-mesh/fusion-runtime` does exactly that. `UnifiedGraphManager` hands `this.opts.additionalResolvers` to every unified-graph build, so a Hive Gateway that reloads its supergraph at runtime (polling, Hive CDN, any dynamic source) re-stitches with the same module-level resolver map on every reload. Each reload adds one wrapper to every stitched field config, and each wrapper keeps its build's `stitchingInfo` reachable: the subschema map, every subschema's `GraphQLSchema`, executors and transports of a generation that `gracefulSchemaReload` has already disposed. Nothing in the old generation can be collected.

What that looks like on a production gateway (2.11.2, a few dozen subgraphs, multi-MB supergraph), post-GC `heapUsed` per reload:

```
216 MB → 350 → 486 → 621 → 757 → 893 → FATAL ERROR: Ineffective mark-compacts near heap limit
```

+135 MB per reload, linear, until Node aborts at the old-space cap; a day with a handful of supergraph publications restarts every instance. Heap snapshots taken between reloads show every schema-shaped constructor growing by exactly one generation (`GraphQLSchema` 468 → 702 across two reloads, one executor closure per subgraph, the supergraph SDL's parse tokens), and the shortest retainer path to every stale `GraphQLSchema` is the same:

```
additionalResolvers.<Type>.<field>.resolve      fn wrappedResolve
  → fn wrappedResolve → fn wrappedResolve       one per reload
    → stitchingInfo → subschemaMap → Subschema → schema / executor / transports
```

Before runtime supergraph reloads, every schema change restarted the process, which is why a leak in a function that is normally called once never surfaced. Any consumer that reuses a `resolvers` object across `stitchSchemas` calls has it, not only fusion-runtime.

### Proposal

Clone the merged resolver map before `addLocalFieldResolvers` runs — the type maps of object and interface types, and their field config objects, one level each — so the wrapper lands on the copy that `addResolversToSchema` consumes and the caller's objects stay untouched. Everything else (enum value maps, scalar instances or configs, union and input type resolvers, functions) passes through by reference, since it is never mutated and enum internal values must keep their identity. The clone must not test for plain objects: `mergeResolvers` deep-merges a type that appears in more than one map into an object whose prototype is not `Object.prototype`, while still reusing the field config objects by reference, and that is precisely the shape a type with both user resolvers and `@resolveTo`-generated ones has. A first attempt with a plain-object guard still leaked through such a type.

PR with the implementation, a regression test that fails on `main` (`expected [Function wrappedResolve] to be [Function resolveOwner]`) and a changeset: https://github.com/graphql-hive/gateway/pull/2591. With the change applied, the same production-shaped gateway holds at 216–219 MB across four reloads.

kpiotr:
Fixes #2592.

## Problem

`stitchSchemas` wraps every user resolver whose field returns a merged type so the merged type's fields get resolved (`addLocalFieldResolvers`). The wrapper is written back into the caller's own field config:

```ts
if (existing != null && typeof existing === 'object') {
  existing.resolve = wrappedResolve;
}
```

`wrappedResolve` closes over `baseResolve` (the previous `existing.resolve`) and this call's `stitchingInfo`. When the same `resolvers` object is passed to `stitchSchemas` again — which is what `@graphql-mesh/fusion-runtime` does with `additionalResolvers` on every supergraph reload — the second call wraps the first call's wrapper, the third wraps the second, and so on. Every wrapper in the chain keeps its call's `stitchingInfo` reachable: the subschema map, every subschema's `GraphQLSchema`, executors and transports of a schema generation that has otherwise been disposed.

We hit this in production on a Hive Gateway (`@graphql-hive/gateway` 2.11.2, stitch 10.2.2) that reloads its supergraph without restarting: heap grew by ~135 MB per reload, linear in the number of reloads, until Node aborted at the V8 heap limit after ~6 reloads. Heap-snapshot retainer paths for every stale `GraphQLSchema` went through `additionalResolvers.<Type>.<field>.resolve → wrappedResolve → wrappedResolve → …`. With this change the heap is flat across reloads (216 → 219 MB over four reloads in the same setup).

## It is also a behaviour bug, not only a leak

`wrappedResolve` closes over *its own* call's `stitchingInfo`, and when the wrappers nest the innermost — oldest — one runs first. So a field whose config is reused hydrates through a superseded call's stitching plan and executors, and once that hydration satisfies the requested selection the newest wrapper finds nothing missing and returns it unchanged.

A test that gives each generation its own subschema data shows it directly. On `main`:

```
call 1: "person-7-from-generation-1"   OK
call 2: "person-7-from-generation-1"   expected "person-7-from-generation-2"
call 3: "person-7-from-generation-1"   expected "person-7-from-generation-3"
```

Every later call keeps answering from generation 1. In a reloading gateway the executors of the superseded generation still point at the same subgraph URLs, so requests do go out and the data looks plausible — the stale part is the plan, which is why this can run for a long time without being noticed.

## Fix

Clone the merged resolver map two levels (type maps and field config objects; functions and custom scalar instances pass through by reference) before `addLocalFieldResolvers` runs, so the wrapper lands on the copy that `addResolversToSchema` consumes and the caller's objects stay untouched.

The clone is limited to object and interface types, the only ones `addLocalFieldResolvers` writes into; enum value maps, scalar instances or configs, union and input type resolvers and functions pass through by reference, so enum internal values keep their identity. It deliberately does not test for plain objects: `mergeResolvers` deep-merges a type that appears in more than one map into an object whose prototype is not `Object.prototype`, while still reusing the field config objects by reference — exactly the case that leaked for us (a type with both user resolvers and `@resolveTo`-generated ones).

## Test

`packages/stitch/tests/resolverMapNotMutated.test.ts` stitches twice with the same `resolvers` (passed as two maps for one type) and asserts the caller's `resolve` function and `selectionSet` are unchanged, an enum value map keeps its object identity, and the stitched field still resolves the merged type on both schemas. It fails on `main` with `expected [Function wrappedResolve] to be [Function resolveOwner]`.

A second case in the same file covers the behaviour above: it stitches three times with one long-lived resolver map, each call over a subschema whose data names its generation, and asserts the delegated field answers from the current one. It fails on `main` with `expected 'person-7-from-generation-1' to be 'person-7-from-generation-2'`.

The full unit project passes with both (`1726 passed | 21 skipped | 2 todo`), as do `check:types`, `check:lint` and `prettier --check`.

Changeset included (`@graphql-tools/stitch`: patch).

### DIFF
--- a/.changeset/stitch-do-not-mutate-resolvers.md
+++ b/.changeset/stitch-do-not-mutate-resolvers.md
@@ -1,0 +1,7 @@
+---
+'@graphql-tools/stitch': patch
+---
+
+`stitchSchemas` no longer mutates the resolver objects passed in `resolvers`.
+
+The wrapper it adds around resolvers of fields that return a merged type (so the merged type's fields get resolved) used to be written back into the caller's field config (`existing.resolve = wrappedResolve`). A resolver map reused across `stitchSchemas` calls — a gateway's additional resolvers on every supergraph reload — therefore gained one wrapper per call, and every wrapper kept its call's `stitchingInfo`, and with it the entire superseded schema, subschemas and executors, reachable: a heap leak of one schema generation per reload. The merged resolver map is now cloned before the wrappers are added.
--- a/.changeset/stitch-do-not-mutate-resolvers.md
+++ b/.changeset/stitch-do-not-mutate-resolvers.md
@@ -4,4 +4,8 @@
 
 `stitchSchemas` no longer mutates the resolver objects passed in `resolvers`.
 
-The wrapper it adds around resolvers of fields that return a merged type (so the merged type's fields get resolved) used to be written back into the caller's field config (`existing.resolve = wrappedResolve`). A resolver map reused across `stitchSchemas` calls — a gateway's additional resolvers on every supergraph reload — therefore gained one wrapper per call, and every wrapper kept its call's `stitchingInfo`, and with it the entire superseded schema, subschemas and executors, reachable: a heap leak of one schema generation per reload. The merged resolver map is now cloned before the wrappers are added.
+The wrapper it adds around resolvers of fields that return a merged type (so the merged type's fields get resolved) used to be written back into the caller's field config (`existing.resolve = wrappedResolve`). A resolver map reused across `stitchSchemas` calls — a gateway's additional resolvers on every supergraph reload — therefore gained one wrapper per call, and every wrapper kept its call's `stitchingInfo`, and with it the entire superseded schema, subschemas and executors, reachable: a heap leak of one schema generation per reload.
+
+Nesting also changed behaviour, because `wrappedResolve` closes over its own call's `stitchingInfo` and the innermost (oldest) wrapper runs first: such a field hydrated through a superseded call's stitching plan and executors, and once that hydration satisfied the selection the newest wrapper found nothing left to resolve.
+
+The merged resolver map is now cloned before the wrappers are added. Enum value maps, scalars, union and input type resolvers and plain resolver functions are still passed through by reference, so their identities are unchanged.
--- a/packages/stitch/src/stitchSchemas.ts
+++ b/packages/stitch/src/stitchSchemas.ts
@@ -103,7 +103,16 @@ export function stitchSchemas<
     } as any);
   }
 
-  const resolverMap: IResolvers = mergeResolvers(resolvers);
+  // addLocalFieldResolvers writes the wrappers it builds back into this map, so
+  // clone it two levels first and leave the caller's resolver objects untouched.
+  // A resolver map reused across stitchSchemas calls (a gateway's additional
+  // resolvers on every schema reload) would otherwise gain one wrapper per call,
+  // each closing over that call's stitchingInfo and keeping the whole superseded
+  // schema alive.
+  const resolverMap: IResolvers = cloneResolverMap(
+    mergeResolvers(resolvers),
+    schema,
+  );
 
   const finalResolvers = inheritResolversFromInterfaces
     ? extendResolversFromInterfaces(schema, resolverMap)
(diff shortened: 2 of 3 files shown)


enisdenjo:
this is looking great! thank you for working on this. I wonder if you think there could be performance implications with this approach? while runining this, have you noticed and impact? thanks 🙏

kpiotr:
No measurable cost, in a benchmark or in production.

The copy is made once per `stitchSchemas` call, and nothing changes on the resolver hot
path: the wrapper is the same function, written to the copy that `addResolversToSchema`
consumes instead of into the caller's object. It is two shallow levels of the merged
resolver map — type maps and field config objects — and only for object and interface
types; functions, scalar instances and enum value maps pass through by reference. So it
allocates one object per type plus one per field config, and that copy dies with the schema
generation that owns it.

Benchmark, this branch against its base (`f8ad168`), same machine, Node 22.13, patched and
base runs alternated twice, median of 31 `stitchSchemas` calls per size. Every field config
returns a merged type, so all of them are wrapped:

| wrapped field configs | base | this PR |
| --- | --- | --- |
| 400 | 3.24 ms / 3.07 ms | 3.07 ms / 3.43 ms |
| 4,000 | 23.48 ms / 23.05 ms | 23.33 ms / 23.54 ms |

Both sizes are inside run-to-run noise and the sign flips between repeats. 4,000 wrapped
field configs is well above the size of the map our own gateway passes in.

In production we have run it as a patch on 10.2.2 since 3 September, on a gateway that
reloads its supergraph in process. Heap is flat across reloads and there have been no
heap-limit aborts since; before it, heap grew by ~135 MB per reload until the abort. CPU
per pod is unchanged.

PR #2614, opened 2026-09-11
Add a dev source for the supergraph

jdolle:
Add a `dev` supergraph "type" that lets the gateway compose a supergraph directly from local/introspected subgraphs, and optionally via remote Hive registry composition, without running a separate [`hive dev`](http://the-guild.dev/graphql/hive/docs/api-reference/cli#local-subgraph-development) process.

This is configurable via a config file or through command line arguments:
  - `--dev-remote` (`DEV_REMOTE`) enables remote composition via the Hive registry. The env var is parsed as a boolean, so `DEV_REMOTE=0` disables it and overrides `remote: true` from the config file.
  - `--dev-registry` (`DEV_REGISTRY`) sets the Hive registry endpoint. It defaults to `https://app.graphql-hive.com/graphql`, so it only needs to be set for self-hosted Hive.
  - The registry target and access token for remote composition are not dev-specific options. They are inherited from the existing global `--hive-target` (`HIVE_TARGET`, as `org/project/target` or a target UUID) and `--hive-access-token` (`HIVE_ACCESS_TOKEN`), so a target and token already configured for usage reporting or tracing are reused. As before, setting both also enables usage reporting. Starting with `--dev-remote` and no token fails fast with an error pointing at `--hive-access-token` (or `token` in the config file).
  - `--dev-service <name>=<url>` (repeated per service) can define the dev source's services
  - `--dev-service-source <name>=federation|graphql|file` specifies the source of the subgraph SDL
    `--dev-service-schema <name>=<path>` is only required if the service source is "file". This specifies the path (from cwd) to the schema file

The reason for the `<name>=...` syntax is that the schema is optional so I had to figure out a way let people mix and match the source type and not have an empty schema argument. So rather than keying by index I added a name to the value. I also considered using JSON but JSON in command line can be awkward.

The CLI option handling lives in `handleDevSupergraphConfig.ts` (same pattern as `handleReportingConfig`) and is covered by unit tests in `packages/gateway/tests/handleDevSupergraphConfig.test.ts`.

An example command when using federated introspection and remote composition:
```
export HIVE_ACCESS_TOKEN=<hive registry access token>
export HIVE_TARGET=<target uuid, or org/project/target>
# only needed for self-hosted Hive
export DEV_REGISTRY=<hive registry endpoint>
```

```sh
gateway supergraph \
  --dev-remote \
  --dev-service "products=http://localhost:8081/graphql"
```

A blog post, product update, and update to documentation is also available at https://github.com/the-guild-org/website/pull/2010


packages/gateway/src/commands/supergraph.ts
```diff
@@ -33,6 +40,100 @@ import { handleFork } from './handleFork';
 import { handleOpenTelemetryCLIOpts } from './handleOpenTelemetryCLIOpts';
 import { handleReportingConfig } from './handleReportingConfig';
 
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
```
jdolle, line 43: also considered not checking this. But it felt preferable to fail on boot for a malformed uuid.

packages/runtime/package.json
```diff
@@ -49,7 +49,7 @@
     "@envelop/disable-introspection": "^9.0.0",
     "@envelop/generic-auth": "^11.0.0",
     "@envelop/instrumentation": "^1.0.0",
-    "@graphql-hive/core": "^0.22.1",
+    "@graphql-hive/core": "0.23.0-alpha-20260910214706-54ac3f3b7bbd1cc0e7a7c4ee7ca116256d0b9d75",
```
jdolle, line 52: will need updated before merging

.changeset/dev-supergraph-source.md
```diff
@@ -0,0 +1,16 @@
...
+gateway compose a supergraph directly from local/introspected subgraphs (optionally via remote
+Hive registry composition) without running a separate `hive dev` process.
+
+Configure it via `supergraph: { type: 'dev', services: [...] }` in the config file, with
+`--dev-remote`, `--dev-registry`, `--dev-registry-token` and `--dev-target` CLI/env overrides for
+the `supergraph` command. The services themselves can also be defined entirely from the CLI using
+`--dev-service-url <name>=<url>` (repeated once per service), with `--dev-service-source
+<name>=federation|graphql|file` and `--dev-service-schema <name>=<path>` as optional per-service
+overlays keyed by the same service name; when used, these take precedence over any `services`
+configured in the config file.
```
n1ru4l, line 16: When I update my graphql schema file or the subgraph serves a new schema, how is that change reflected?

E.g.
- do we watch file events?
- do we poll each subgraph and cause a new composition in case the subgraph schema changed?

jdolle: Because this is a fetcher built into the gateway, the gateway handles polling to refetch the schema.
So there's no watcher logic required.
The subgraph is read and checked if it has changed. If so, a new composition is ran.

Also, the composition call is also rate limited to prevent calling the composer too often.

yarn.lock
```diff
@@ -4581,26 +4581,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@graphql-hive/core@npm:0.21.1-alpha-20260604213042-0d6112116193a921a379bbc3b917702c29368479":
```
jdolle, line 4584: Did we intend this to be an alpha version?

packages/runtime/src/fetchers/dev.ts
```diff
@@ -0,0 +1,433 @@
...
+
+function servicesUnchanged(previous: Service[], next: Service[]): boolean {
+  if (previous.length !== next.length) {
+    return false;
+  }
+
+  return next.every(
+    (service) =>
+      previous.find((p) => p.name === service.name)?.sdl === service.sdl,
+  );
```
enisdenjo: cache equality ignores the service URL, even though that URL is embedded in the composed supergraph and controls request routing.

changing `products` from one endpoint to another with unchanged SDL (like url overrides) returns the cached supergraph pointing at the old endpoint.

I'd rather compare both `url` and `sdl`. what do you think?

packages/runtime/src/fetchers/dev.ts
```diff
@@ -0,0 +1,433 @@
...
+      // nothing listens to the `snapshot` event, so don't schedule its interval
+      enableSnapshots: false,
+    },
+  );
+
+  return {
+    async fetch() {
+      const services = await resolveServices(devOpts.services, cwd, fetch);
+
+      const cached: CachedSupergraph | undefined = await cache?.get(CACHE_KEY);
```
enisdenjo: here it's always the same cache key, do you think that maybe we should also add the configuration?

I am thinking of such a case:

1. gateway composes locally and stores the result
2. configuration is changed to remote composition against a Hive target
3. cache survives the restart, such as when Redis is used
4. local service schemas have not changed
5. fetcher returns the previous local result without contacting Hive

packages/gateway/src/commands/supergraph.ts
```diff
@@ -65,6 +173,61 @@ export const addCommand: AddCommand = (ctx, cli) =>
         'env',
       );
     })
+    .addOption(
+      new Option(
+        '--dev-remote',
+        'Compose the dev supergraph remotely via the Hive registry instead of composing locally. Only applies when the supergraph source is a Hive dev fetcher (`supergraph: { type: "dev", ... }` in the config file).',
+      ).env('DEV_REMOTE'),
```
enisdenjo, line 76: commander treats a boolean environment option as enabled when the variable exists, it does not interpret its value as a boolean!

that means both of these enable remote mode:

```
DEV_REMOTE=1
DEV_REMOTE=0
```

see:

https://github.com/graphql-hive/gateway/blob/c0e1c7dff4432eb0483de0d89d10ba506ecaa819/packages/gateway/src/commands/supergraph.ts#L52-L67

enisdenjo (changes requested):
I like that it's just another fetcher, but you say:

> Add a dev supergraph "type", backed by @graphql-hive/core's createDevFetcher, that lets the gateway compose a supergraph directly from local/introspected subgraphs

however, I dont see the createDevFetcher comes from the hive/core package?

---

you atm have to:

```
supergraph: {
  type: 'dev',
  remote: true,
  token: process.env.HIVE_TOKEN,
  target: {
    byId: 'target-uuid',
  },
  services: [
    {
      name: 'products',
      url: 'http://localhost:8081/graphql',
    },
  ],
}
```

can we instead use the hive token from env? HIVE_CDN_KEY or HIVE_ACCESS_KEY? we already use it for other hive things, and doing it here too will keep things in line

---

otherwise, great work!

packages/runtime/src/fetchers/dev.ts
jdolle: 💯

packages/runtime/src/fetchers/dev.ts
jdolle: With my current logic, we could always have a case where the remote schema changes and unless you're modifying the local subgraph, it wont refetch. This was intentional since this is meant to be used for active development and I thought this struck a good balance with the number of fetches.

The case you outlined is a concern though, since at boot you'd expect it to run the latest on boot.

I've added logic to ensure it always fetches the latest on the first run.

jdolle:
> I dont see the createDevFetcher comes from the hive/core package?

Fixed description. I had originally created the fetcher in hive/core, but moved it after getting feedback from Laurin to keep the dependencies cleaner.

> can we instead use the hive token from env? HIVE_CDN_KEY or HIVE_ACCESS_KEY? we already use it for other hive things, and doing it here too will keep things in line

We can. I've adjusted this.

---

I believe all comments have been addressed.

yarn.lock
enisdenjo: nope

packages/runtime/src/fetchers/dev.ts
enisdenjo: always composing on boot fixes stale entries from a previous run, but live gateways sharing Redis can still overwrite and consume each otherss cached composition when their service URLs and SDLs match but their composition mode, registry, or target differ, right?

I pushed a follow-up that keeps the composed supergraph in the fetcher instance instead of the shared cache. this still composes once on boot and only recomposes when a local SDL or URL changes, while preventing gateways with different composition configurations from consuming each other’s cached supergraphs.

https://github.com/graphql-hive/gateway/pull/2614/commits/894ae86553061ec1531db1d67419d58562d0a06f

enisdenjo (approved):

n1ru4l (approved):

